PR #1, opened 2020-08-04
Refactored math

albertocuestacanada:
Check in the code for a line-by-line description of the changes.

Tests to be fixed, shouldn't be hard.

### DIFF
--- a/contracts/USM.sol
+++ b/contracts/USM.sol
@@ -30,14 +30,14 @@ contract USM is ERC20 {
     /**
      * @notice Multiplies x and y, assuming they are both fixed point with 27 digits.
      */
-    function muld(uint256 x, uint256 y) internal pure returns (uint256) {
+    function mulFixed(uint256 x, uint256 y) internal pure returns (uint256) {
         return x.mul(y).div(UNIT);
     }
 
     /**
      * @notice Divides x by y, assuming they are both fixed point with 27 digits.
      */
-    function divd(uint256 x, uint256 y) internal pure returns (uint256) {
+    function divFixed(uint256 x, uint256 y) internal pure returns (uint256) {
         return x.mul(UNIT).div(y);
     }
 
@@ -46,7 +46,7 @@ contract USM is ERC20 {
      */
     function mint() external payable {
         uint usmAmount = _ethToUsm(msg.value);
-        uint usmMinusFee = usmAmount.sub(muld(usmAmount, MINT_FEE));
+        uint usmMinusFee = usmAmount.sub(mulFixed(usmAmount, MINT_FEE));
         ethPool = ethPool.add(msg.value);
         _mint(msg.sender, usmMinusFee);
     }
@@ -58,7 +58,7 @@ contract USM is ERC20 {
      */
     function burn(uint _usmAmount) external {
         uint ethAmount = _usmToEth(_usmAmount);
-        uint ethMinusFee = ethAmount.sub(muld(ethAmount, BURN_FEE));
+        uint ethMinusFee = ethAmount.sub(mulFixed(ethAmount, BURN_FEE));
         ethPool = ethPool.sub(ethMinusFee);
         _burn(msg.sender, _usmAmount);
         Address.sendValue(msg.sender, ethMinusFee);
@@ -74,7 +74,7 @@ contract USM is ERC20 {
         if (ethPool == 0) {
             return 0;
         }
-        return divd(totalSupply(), _ethToUsm(ethPool));
+        return divFixed(totalSupply(), _ethToUsm(ethPool));
     }
 
     /**
@@ -86,7 +86,7 @@ contract USM is ERC20 {
      */
     function _ethToUsm(uint _ethAmount) internal view returns (uint) {
         require(_ethAmount > 0, "Eth Amount must be more than 0");
-        return muld(_oraclePrice(), _ethAmount);
+        return mulFixed(_oraclePrice(), _ethAmount);
     }
 
     /**
@@ -98,7 +98,7 @@ contract USM is ERC20 {
      */
     function _usmToEth(uint _usmAmount) internal view returns (uint) {
         require(_usmAmount > 0, "USM Amount must be more than 0");
-        return divd(_usmAmount, _oraclePrice());
+        return divFixed(_usmAmount, _oraclePrice());
     }
 
     /**
--- a/contracts/USM.sol
+++ b/contracts/USM.sol
@@ -12,11 +12,9 @@ import "./IOracle.sol";
 contract USM is ERC20 {
     using SafeMath for uint;
 
-    // 0.1% (10 parts per 10,000)
-    uint constant MINT_FEE = 10;
-    // 0.5%; (50 parts per 10,000)
-    uint constant BURN_FEE = 50;
-    uint constant PERCENTAGE_BASE = 10000;
+    uint constant UNIT = 1000000000000000000000000000; // 10**27, MakerDAO's RAY
+    uint constant MINT_FEE = UNIT / 1000; // 0.1%
+    uint constant BURN_FEE = UNIT / 200; // 0.5%
 
     address oracle;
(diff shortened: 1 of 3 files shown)


contracts/USM.sol
```diff
@@ -12,11 +12,9 @@ import "./IOracle.sol";
 contract USM is ERC20 {
     using SafeMath for uint;
 
-    // 0.1% (10 parts per 10,000)
-    uint constant MINT_FEE = 10;
-    // 0.5%; (50 parts per 10,000)
-    uint constant BURN_FEE = 50;
-    uint constant PERCENTAGE_BASE = 10000;
+    uint constant UNIT = 1000000000000000000000000000; // 10**27, MakerDAO's RAY
```
albertocuestacanada, line 15: It's best to use a single unit for decimal numbers across the project. Remember that token amounts are actually integers (wei) even if the ERC20 standard has a `decimals()` variable for convenience.

contracts/USM.sol
```diff
@@ -12,11 +12,9 @@ import "./IOracle.sol";
...
+    uint constant MINT_FEE = UNIT / 1000; // 0.1%
```
albertocuestacanada, line 16: In UNIT, both of them.

contracts/USM.sol
```diff
@@ -29,13 +27,28 @@ contract USM is ERC20 {
         ethPool = 0;
     }
 
+    /**
```
albertocuestacanada, line 30: These functions are actually `integer * decimal = integer` operations, assuming that the decimal is in UNIT

contracts/USM.sol
```diff
@@ -29,13 +27,28 @@ contract USM is ERC20 {
...
+     */
+    function divd(uint256 x, uint256 y) internal pure returns (uint256) {
+        return x.mul(UNIT).div(y);
+    }
+
     /**
      * @notice Mint ETH for USM. Uses msg.value as the ETH deposit.
      */
     function mint() external payable {
-        uint usmAmount = _applyFee(_ethToUsm(msg.value), MINT_FEE);
```
albertocuestacanada, line 36: Removing `applyFee` to make operations a bit more explicit.

contracts/USM.sol
```diff
@@ -44,52 +57,24 @@ contract USM is ERC20 {
...
-     * @param _number The amount that the fee is being applied to
-     * @param _feePercentage Percentage fee to be subtracted from the number
-     * in parts per 10,000.
-     * @return The result of the _number minus the percentage fee.
-     */
-    function _applyFee(uint _number, uint _feePercentage) internal pure returns (uint) {
-        uint result = _number.sub((_number.mul(_feePercentage)).div(PERCENTAGE_BASE));
-        require(result < _number, "Fee must make the result smaller");
-        return result;
+        return divd(totalSupply().mul(UNIT), _ethToUsm(ethPool)); // This means a pool upper limit of about 10**50 wei.
```
albertocuestacanada, line 77: `debtRatio` is reduced to a one-liner. Since both operands are `integer` and we want the result to be `decimal`, we multiply the first one by `UNIT`

contracts/USM.sol
```diff
@@ -114,20 +98,15 @@ contract USM is ERC20 {
...
      *
-     * @return (latest price, decimal shift)
+     * @return price in UNIT
      */
-    function _oracleData() internal view returns (uint, uint) {
-        return (
-            IOracle(oracle).latestPrice(),
-            IOracle(oracle).decimalShift()
-        );
+    function _oraclePrice() internal view returns (uint) {
```
albertocuestacanada, line 109: We can make the oracle return a price in a UNIT format directly usable by the rest of the contract.

contracts/USM.sol
```diff
@@ -44,52 +57,24 @@ contract USM is ERC20 {
...
-     * percentage fee.
-     *
-     * @param _number The amount that the fee is being applied to
-     * @param _feePercentage Percentage fee to be subtracted from the number
-     * in parts per 10,000.
-     * @return The result of the _number minus the percentage fee.
-     */
-    function _applyFee(uint _number, uint _feePercentage) internal pure returns (uint) {
-        uint result = _number.sub((_number.mul(_feePercentage)).div(PERCENTAGE_BASE));
-        require(result < _number, "Fee must make the result smaller");
```
albertocuestacanada, line 91: These `require` are not needed, because both parameters are `uint` and therefore positive, and `.sub` protects us against overflow.

contracts/USM.sol
```diff
@@ -29,13 +27,28 @@ contract USM is ERC20 {
         ethPool = 0;
     }
 
+    /**
+     * @notice Multiplies x and y, assuming they are both fixed point with 27 digits.
+     */
+    function muld(uint256 x, uint256 y) internal pure returns (uint256) {
```
alexroan, line 33: Could we make `muld` and `divd` a little more descriptive?

Perhaps something like `mulRay` and `divRay`? (not entirely convinced by these either 😂 )

I'm open to ideas on this one.

contracts/USM.sol
```diff
@@ -29,13 +27,28 @@ contract USM is ERC20 {
...
+        return x.mul(y).div(UNIT);
+    }
+
+    /**
+     * @notice Divides x between y, assuming they are both fixed point with 27 digits.
```
alexroan, line 38: Change to: `@notice Divides x by y, assuming they are both fixed point with 27 digits.`

contracts/USM.sol
alexroan: I think I'm understanding this right... If I am it changes the current `debtRatio` from returning a below `1` value when overcollateralized to a below `100` value. For example, what was 0.9, is now 90 (as in 90%)?

Using smaller numbers (without the constraints of solidity in this thought example), let's assume the following:

* `totalSupply` in USM = 100
* `_ethToUsm(ethPool)` in USM = 110

.. technically we are slightly overcollateralized for whatever reason, which is where we want to be.

Let's assume `UNIT` = 10 (purely for this thought example).

Running this function would perform the following:

1. `totalSupply().mul(UNIT)` = 100 * 10 = 1,000
2. `x.mul(UNIT).div(y)` = (1,000 * 10) / 110 = 10,000 / 110 = 90.909090....

~90.91% debt ratio, which looks right!

I believe that in the master branch, the result would be ~0.909.

Am I getting that right?

albertocuestacanada: Before `debtRatio` was returning a fixed point number with `decimals()` decimals, now it is returning a fixed point number with 27 decimals, which is the format for all fixed point numbers in the contract.

I think that to fix the tests, I just need to add 9 zeros to each one of your `debtRatio` tests.

contracts/USM.sol
albertocuestacanada: I've used `muld` (for Multiply Decimals) for ages, but I'm not emotionally attached to it. `mulf` (for Multiply Fixed point numbers) would be probably better. If you want to go for longer function names I would do `mulFixed` or `mulF27` over `mulRay`.

contracts/USM.sol
albertocuestacanada: Me English no good :P

contracts/USM.sol
alexroan: I like `mulFixed`!

albertocuestacanada:
Fixed the tests. Cross-type operations are a bit weird.

```
mulFixed:
RAY * RAY = output in RAY
RAY * UINT = output in UINT
UINT * RAY = output in UINT
UINT * UINT = bad output

divFixed:
RAY / RAY = output in RAY
RAY / UINT = bad output
UINT / RAY = output in UINT
UINT / UINT = output in RAY
```
Usually, only UINT * RAY and UINT / RAY are used. I only realized the other cases because one conversion was getting me confused. Type-checked decimal math would indeed be nice.

albertocuestacanada:
Also, we will need a function in the tests to convert from human readable (0.999) to RAY (999000000000000000000000000) and vice versa.

PR #2, opened 2020-08-17
Installed buidler and other tools

albertocuestacanada:
Run `yarn` and then `yarn test` (or `npx buidler test`).

Check the `console.log` calls in `USM.sol`.

For the same price, I got gas-reporter and solidity-coverage running. Not so sure about linting.

The one thing left to do is to set up network deployments, there is a buidler-deploy plugin that seems to come with cool new features each day.


alexroan:
Love this!

PR #4, opened 2020-08-29
Token/fum

albertocuestacanada:
TBA


contracts/USM.sol
```diff
@@ -1,57 +1,48 @@
...
  */
-contract USM is ERC20 {
+contract USM is BufferedToken {
     using SafeMath for uint;
+    using WadMath for uint;
 
-    uint constant UNIT = 1e27; // 10**27, MakerDAO's RAY
-    uint constant MINT_FEE = UNIT / 1000; // 0.1%
-    uint constant BURN_FEE = UNIT / 200; // 0.5%
+    uint constant MINT_FEE = WAD / 1000; // 0.1%
```
albertocuestacanada, line 18: The comments should say `// 0.001 ETH` and `// 0.005 ETH` since these are not percentages anymore.

contracts/USM.sol
```diff
@@ -1,57 +1,48 @@
...
+    uint constant BURN_FEE = WAD / 200; // 0.5%
+    uint constant MAX_DEBT_RATIO = 900000000000000000; // 90%
```
albertocuestacanada, line 20: You can use scientific notation and write it as ` = 9e17`, or stick to WAD and do ` = WAD * 9/10`

jacob-eliosoff:
Just starting to look here, thoughts before I dig into the actual logic:

1. You might consider just scrapping all the `MINT_FEE`/`BURN_FEE` stuff if it makes life simpler (probably), because I think the const-product fees are likely to be all we use in the end.  I could submit a PR.

2. I read a bit about wads & rays...  So Eth's fixed-point convention is to shift by decimal places, not binary?  (Eg in binary `MAX_DEBT_RATIO` might be 90% * 2^64 = 16602069666338596454, rather than 90% * 10^18 = 900000000000000000.)  I'd have thought SHL/SHR make binary a little more appealing but probably just do what's normal...

3. Is your `latestFumPrice` in terms of $ or ETH?  Quite important that it's ETH, ie, moves up & down with ETH/USD.  (That is, in-ETH may or may not be the best way, but whether it's $ or ETH is def important and in [my Python proof-of-concept](https://github.com/jacob-eliosoff/usm-stablecoin) it's ETH.)

4. Rethinking what I wrote about the "big ETH price moves since the last call" issue - I think all four ops, eg `mint`, should just proceed like this, keeping track of the old debt ratio rather than the old ETH price or FUM price:

```
function mint() external payable {
    _updateLatestEthPrice()                                 // Or whatever - from the oracle
    _updateMinFumBuyPrice()
    _mint()
}

function _updateMinFumBuyPrice(uint _fumPriceInEth) internal {
    uint currentDebtRatio = debtRatio()
    if (currentDebtRatio <= MAX_DEBT_RATIO) {               // We've dropped below max dr
	_minFumBuyPriceInEth = 0                            // Clear mfbp
    } else if (_previousDebtRatio <= MAX_DEBT_RATIO) {      // We were < max dr, but have now crossed above - so set mfbp
	// This calculates the mfbp, in ETH terms, at the point we crossed max dr.  Note how few vars it depends on:
	_minFumBuyPriceInEth = (1 - MAX_DEBT_RATIO) * ethPool / FUM(fum).totalSupply()
    }
    _previousDebtRatio = currentDebtRatio
}
```

jacob-eliosoff:
Actually, we don't even need to store `_previousDebtRatio`.  Just the mfbp.  Like this:

```
function mint() external payable {
    _updateLatestEthPrice()                             // Or whatever - from the oracle
    _updateMinFumBuyPrice()
    _mint()
}

function _updateMinFumBuyPrice(uint _fumPriceInEth) internal {
    if (debtRatio() <= MAX_DEBT_RATIO) {                // We've dropped below max dr (or were already below it, whatev)
	_minFumBuyPriceInEth = 0                        // Clear mfbp
    } else if (_minFumBuyPriceInEth == 0) {             // We were < max dr, but have now crossed above - so set mfbp
	// This calculates the mfbp, in ETH terms, at the point we crossed max dr.  Note how few vars it depends on:
	_minFumBuyPriceInEth = (1 - MAX_DEBT_RATIO) * ethPool / FUM(fum).totalSupply()
    }
}
```

jacob-eliosoff:
Also, we're all just pulling numbers from the air but I lean towards `MAX_DEBT_RATIO` = 80% rather than 90%.  It's a pretty big difference because 90% means that if the system goes underwater, new FUM buyers are paying half the price (total FUM valuation = 10% of pool) as they are if we set max to 80% (FUM valuation = 20% of pool).  So the tradeoff is, a lower, "more conservative" max will trigger a warning state (eg, FUM redemptions paused) more frequently, perhaps worsening UX for USM users; but those warning states will be less hazardous - in particular, less likely to dilute/wipe out existing FUM holders.

contracts/USM.sol
albertocuestacanada: > From @jacob-eliosoff: I read a bit about wads & rays... So Eth's fixed-point convention is to shift by decimal places, not binary? (Eg in binary MAX_DEBT_RATIO might be 90% * 2^64 = 16602069666338596454, rather than 90% * 10^18 = 900000000000000000.) I'd have thought SHL/SHR make binary a little more appealing but probably just do what's normal... 

Shift operators in solidity work on the binary representation of the number.
Which usually is written in hexadecimal as 0x...
And which is stored in a uint anyway.
uint x = 10; x << 1; returns 20
uint x = 0xa; x << 1; returns 20
So you can define your numbers in decimal, so that you know what you are typing, and then use bit shift operators if you want to speed up sqaure and square root operations

jacob-eliosoff: FYI, I [updated](https://github.com/jacob-eliosoff/usm-stablecoin/commit/e824c238fdfe6788d92519421bd0f2b694a35a47) usm_constproduct.py's half_exp_approx() to use decimal-shifted inputs rather than binary-shifted - it doesn't affect the algorithm.

albertocuestacanada: Cool! Just as a question, would it be very hard to convert this from a recursive to an iterative function?

jacob-eliosoff: Ha yeah I wavered about that when writing it.  I had done some Googling that suggested recursion in Solidity might not actually be too expensive?  But anyway, as discussed in chat, the right thing here is probably to lean on the `rpow()` you found in ds-math, which basically _is_ the iterative version of the algo I used.  `rpow(x, n)` has the flaw for our purposes that it assumes `n` is an integer, whereas we need to handle fractional exponents, but we can get around that:

```
HALF_TO_THE_TENTH = 933032991536807415981343266

function rhalfexp(uint x) returns (uint z) {                // Approximates 0.5**x
    return rpow(HALF_TO_THE_TENTH, (10 * x + RAY / 2) / RAY)
}
```

Or we could use HALF_TO_THE_HUNDREDTH to get 0.01 input precision in exchange for ~three more iterations inside `rpow()`.  But we don't really need more than 0.1 input precision, the above is probably fine.

albertocuestacanada: I've never actually researched recursion in Solidity, so let's give it a go and see what happens!

jacob-eliosoff: Nah the thing is `rpow()` makes my recursion obsolete!  Recursion might come up again though...

contracts/USM.sol
```diff
@@ -1,117 +1,131 @@
...
+            return latestFumPrice;
+        }
 
-    /**
-     * @notice Divides x by y, assuming they are both fixed point with 27 digits.
-     */
-    function divFixed(uint256 x, uint256 y) internal pure returns (uint256) {
-        return x.mul(UNIT).div(y);
+        uint fumTotalSupply = fum.totalSupply().max(WAD); // fumTotalSupply is floored at WAD
+        return uint(buffer).wadDiv(fumTotalSupply);
```
jacob-eliosoff, line 50: I suggest something like:

```
theoreticalFumPrice = uint(buffer).wadDiv(fumTotalSupply);
minFumPrice = _ethToUsm(_minFumBuyPriceInEth);
if (minFumPrice > theoreticalFumPrice) {    // Does Solidity have a max() fn?  Maybe we should write one...
    return minFumPrice;
} else {
    return theoreticalFumPrice;
}

File: contracts/USM.sol [hunk 4]
```diff
@@ -1,117 +1,131 @@
...
-    function burn(uint _usmAmount) external {
-        uint ethAmount = _usmToEth(_usmAmount);
-        uint ethMinusFee = ethAmount.sub(mulFixed(ethAmount, BURN_FEE));
-        ethPool = ethPool.sub(ethMinusFee);
-        _burn(msg.sender, _usmAmount);
-        Address.sendValue(msg.sender, ethMinusFee);
+    function burn(uint _usmToBurn) external {
+        require(_usmToBurn >= MIN_BURN_AMOUNT, "Must burn at least 1 USM");
+        uint ethToSend = _burn(_usmToBurn);
+        require(debtRatio() <= MAX_DEBT_RATIO,
```
jacob-eliosoff, line 71: I suggest `require(debtRatio() <= WAD)`, ie, 100% rather than `MAX_DEBT_RATIO` = (eg) 80%; because when we're between 80% and 100%, `burns` actually _decrease_ debt ratio!  Eg, suppose price is $400, the pool contains 100 ETH (total value: $40,000), and there are 36,000 USM outstanding, so debt ratio is $36,000 / $40,000 = 90%.  And now suppose a user burns 1,000 USM, removing 2.5 ETH.  Then the new debt ratio is $35,000 / $39,000 = 89.74%.

contracts/WadMath.sol
```diff
@@ -0,0 +1,20 @@
...
+    using SafeMath for uint;
+
+    uint constant WAD = 10 ** 18;
+
+    //rounds to zero if x*y < WAD / 2
+    function wadMul(uint x, uint y) internal pure returns (uint) {
+        return ((x.mul(y)).add(WAD.div(2))).div(WAD);
+    }
+
+    //rounds to zero if x*y < WAD / 2
```
jacob-eliosoff, line 15: Correction: `if x/y < WAD / 2`

contracts/BufferedToken.sol
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * @notice Mint USM from Eth. Inheriting contract must handle the Ether transfers.
+     */
+    function _mint(uint ethAmount) internal returns (uint) {
+        uint usmAmount = _ethToUsm(ethAmount);
+        ethPool = ethPool.add(ethAmount);
+        super._mint(msg.sender, usmAmount);
+        return usmAmount;
+    }
```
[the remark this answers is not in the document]
albertocuestacanada: Done

contracts/BufferedToken.sol
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * @notice Burn USM for Eth. Inheriting contract must handle the Ether transfers.
+     */
+    function _burn(uint usmAmount) internal returns (uint) {
+        uint ethAmount = _usmToEth(usmAmount);
+        ethPool = ethPool.sub(ethAmount);
+        super._burn(msg.sender, usmAmount);
+        return ethAmount;
+    }
```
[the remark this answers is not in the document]
albertocuestacanada: Done

contracts/BufferedToken.sol
```diff
@@ -0,0 +1,109 @@
...
+import "./oracles/IOracle.sol";
+
+/**
+ * @title Buffered Token
+ * @author Alex Roan (@alexroan)
+ * @notice This contract represents a token which accepts ETH as collateral to mint and burn ERC20 tokens.
+ * It's functions abstract all oracle, conversion, debt ratio and ETH buffer calculations, so that
+ * inheriting contracts can focus on their minting and burning algorithms.
+ */
+contract BufferedToken is ERC20 {
```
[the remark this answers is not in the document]
albertocuestacanada: The original purpose from @alexroan was to separate the code and make it more manageable. It worked well in that sense for testing.

As it stands, `BufferedToken` is a general buffered token, we are choosing the underlying to be Eth, but the code could be reused to have anything else as the underlying.

We can choose to merge them, but I would probably wait just until we have tests for `fund` and `defund`.

contracts/USM.sol
```diff
@@ -1,117 +1,131 @@
...
  */
-contract USM is ERC20 {
+contract USM is BufferedToken {
     using SafeMath for uint;
+    using Math for uint;
+    using WadMath for uint;
+
+    uint constant public MIN_ETH_AMOUNT = WAD / 1000;         // 0.001 ETH
+    uint constant public MIN_BURN_AMOUNT = WAD;               // 1 USM
+    uint constant public MAX_DEBT_RATIO = 900000000000000000; // 90%
```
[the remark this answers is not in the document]
albertocuestacanada: Done

contracts/BufferedToken.sol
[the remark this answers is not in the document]
alexroan: Yeah, it's for the separation of concerns.  `EthPool`, `debtRatio()`, oracle conversions, etc are managed by BufferedToken so that USM could focus solely on the mathematics of `mint`, `burn`, `fund` and `defund`.

I'm hesitant to merge the two.

alexroan:
> Just starting to look here, thoughts before I dig into the actual logic:
> 
> 1. You might consider just scrapping all the `MINT_FEE`/`BURN_FEE` stuff if it makes life simpler (probably), because I think the const-product fees are likely to be all we use in the end.  I could submit a PR.
> 2. I read a bit about wads & rays...  So Eth's fixed-point convention is to shift by decimal places, not binary?  (Eg in binary `MAX_DEBT_RATIO` might be 90% * 2^64 = 16602069666338596454, rather than 90% * 10^18 = 900000000000000000.)  I'd have thought SHL/SHR make binary a little more appealing but probably just do what's normal...
> 3. Is your `latestFumPrice` in terms of $ or ETH?  Quite important that it's ETH, ie, moves up & down with ETH/USD.  (That is, in-ETH may or may not be the best way, but whether it's $ or ETH is def important and in [my Python proof-of-concept](https://github.com/jacob-eliosoff/usm-stablecoin) it's ETH.)
> 4. Rethinking what I wrote about the "big ETH price moves since the last call" issue - I think all four ops, eg `mint`, should just proceed like this, keeping track of the old debt ratio rather than the old ETH price or FUM price:
> 
> ```
> function mint() external payable {
>     _updateLatestEthPrice()                                 // Or whatever - from the oracle
>     _updateMinFumBuyPrice()
>     _mint()
> }
> 
> function _updateMinFumBuyPrice(uint _fumPriceInEth) internal {
>     uint currentDebtRatio = debtRatio()
>     if (currentDebtRatio <= MAX_DEBT_RATIO) {               // We've dropped below (or already were below) max dr
> 	_minFumBuyPriceInEth = 0                            // Clear mfbp
>     } else if (_previousDebtRatio <= MAX_DEBT_RATIO) {      // We were < max dr, but have now crossed above - so set mfbp
> 	// This calculates the mfbp, in ETH terms, at the point we crossed max dr.  Note how few vars it depends on:
> 	_minFumBuyPriceInEth = (1 - MAX_DEBT_RATIO) * ethPool / FUM(fum).totalSupply()
>     }
>     _previousDebtRatio = currentDebtRatio
> }
> ```

1. We've removed these in this branch 👍 

2. Yep, shifting decimals is the norm

3. `latestFumPrice` is in ETH, yes.

4. What advantages would we get from using `_previousDebtRatio` over the current setup? The reason I opted for `latestFumPrice` was due to simplicity when introducing the FUM token in this branch. The functions were getting quite long and complex so I instead chose `latestFumPrice` so we could iterate on it later. It definitely needs work!

alexroan:
> Also, we're all just pulling numbers from the air but I lean towards `MAX_DEBT_RATIO` = 80% rather than 90%. It's a pretty big difference because 90% means that if the system goes underwater, new FUM buyers are paying half the price (total FUM valuation = 10% of pool) as they are if we set max to 80% (FUM valuation = 20% of pool). So the tradeoff is, a lower, "more conservative" max will trigger a warning state (eg, FUM redemptions paused) more frequently, perhaps worsening UX for USM users; but those warning states will be less hazardous - in particular, less likely to dilute/wipe out existing FUM holders.

Resolved :)

contracts/USM.sol
alexroan: We are essentially doing that a few lines above but with `latestFumPrice` instead of `_minFumBuyPrice`.

contracts/USM.sol
alexroan: Aaaahhhh 😲  this is interesting. Which functions increase, and which functions decrease `debtRatio`?

albertocuestacanada: Fixed :+1:

contracts/BufferedToken.sol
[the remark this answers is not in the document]
albertocuestacanada: I'm now starting to test `fund()`, and I don't think the split works, actually. Both `fund()` and `defund()` manipulate `ethPool`, which is a `BufferedToken` variable. It's not great to manipulate parent variables from inheriting classes.

I'll try to finish the testing as-is, then we merge this PR into `master`, and then I think we should merge `BufferedToken` and `USM` into a single contract (in a separate PR), separating concerns just by using internal and public functions (business logic + api).

albertocuestacanada:
Could someone have a look at [this test](https://github.com/alexroan/USM/blob/d0776d9df2bb6e3828f4e633b1e854856589dc87/test/03_USM.test.js#L122) and find out what's wrong?

 - Shouldn't `ethPool` be 1 WAD, am I getting mocha wrong?
 - What should `fumPrice()` be after doubling the size of the `ethPool` via `fund()`?

jacob-eliosoff:
Thanks for looking Alex.

> 3. `latestFumPrice` is in ETH, yes.

OK, great!

> 4. What advantages would we get from using `_previousDebtRatio` over the current setup? The reason I opted for `latestFumPrice` was due to simplicity when introducing the FUM token in this branch. The functions were getting quite long and complex so I instead chose `latestFumPrice` so we could iterate on it later. It definitely needs work!

See my [follow-up comment](https://github.com/alexroan/USM/pull/4#issuecomment-683363873) - I actually don't think we need to store anything historical other than `minFumBuyPriceInEth` (or whatever we want to call it, but probably not "latest" - `minFumPrice` would do as long as we remember it's in ETH not USD).  Basically an important part of my algo (though there may be better algos) is that the FUM buy price in ETH is set when the system crosses over `MAX_DEBT_RATIO`, and then _not_ updated again until debt ratio drops back below max.  So we should only calculate & store it once, at that crossing point.

jacob-eliosoff:
> Could someone have a look at [this test](https://github.com/alexroan/USM/blob/d0776d9df2bb6e3828f4e633b1e854856589dc87/test/03_USM.test.js#L122) and find out what's wrong?
> 
> * Shouldn't `ethPool` be 1 WAD, am I getting mocha wrong?
> * What should `fumPrice()` be after doubling the size of the `ethPool` via `fund()`?

Happy to have a look...  What's the easiest way for me to run these tests and see their output?  Meanwhile I'll try peering at the code...

contracts/USM.sol
jacob-eliosoff: I see your point, but in fact (again, per my possibly-improvable algo) we need to compare theoretical price to min price, not just check whether we're underwater.  Because there are cases where although we're still underwater, the theoretical price climbs **higher** than the min price, so buyers should pay the former.  Example:

1. 100 ETH @ $400 ($40,000), 30,000 USM (buffer $10,000, dr 75%), 5,000 FUM @ $2
2. Price $400 -> $200.  Debt ratio crossed 80%, mfbp calculated as: (1-80%) * 100 / 5,000 = 0.004 ETH:
100 ETH @ $200 ($20,000), 30,000 USM (buffer -$10,000, dr 150%), 5,000 FUM @ -$2 (min 0.004 ETH = $0.80)
3. Fund 50:
150 ETH @ $200 ($30,000), 30,000 USM (buffer $0, dr 100%), 5,000 FUM @ $0 (min 0.004 ETH = $0.80)
4. Price $200 -> $240:
150 ETH @ $240 ($36,000), 30,000 USM (buffer $6,000, dr 83.33%), 5,000 FUM @ **$1.20** (min 0.004 ETH = **$0.96**)

test/03_USM.test.js
```diff
@@ -0,0 +1,310 @@
...
+                    const usmBalance = (await usm.balanceOf(user)).toString();
+                    const fumPrice = (await usm.latestFumPrice()).toString();
+    
+                    await usm.burn(usmBalance, { from: user });
+                    const newUsmBalance = (await usm.balanceOf(user)).toString();
+                    newUsmBalance.should.equal('0');
+                    // TODO: Test the eth balance just went up. Try setting gas price to zero for an exact amount
+    
+                    const newFumPrice = (await usm.latestFumPrice()).toString();
+                    newFumPrice.should.equal(fumPrice); // Burning doesn't change the fum price if buffer is 0
```
jacob-eliosoff, line 97: USM `mints` and `burns` should never move the FUM price at all, except two cases:
1. If the system is collecting fees, they go to FUM holders, so will (slightly) increase the FUM price.
2. As long as debt ratio ≤ 100%, `burns` decrease debt ratio.  (If debt ratio > 100%, `burns` should be disallowed.)  So it's possible for a `burn` to pull debt ratio < max, causing `minFumBuyPrice` to be cleared and changing the effective FUM buy price back to its theoretical value (buffer / FUM qty) - which might be lower than the previous `minFumBuyPrice` (or might not).

jacob-eliosoff: (Similarly, fees and crossings of `MAX_DEBT_RATIO` aside, `funds` and `defunds` shouldn't move FUM price either!  It mainly just moves when ETH price does.)

jacob-eliosoff:
> Could someone have a look at [this test](https://github.com/alexroan/USM/blob/d0776d9df2bb6e3828f4e633b1e854856589dc87/test/03_USM.test.js#L122) and find out what's wrong?
>
> * Shouldn't `ethPool` be 1 WAD, am I getting mocha wrong?

I had a look...  It does seem like it should be 1 WAD, unless the `beforeEach()` isn't doing what one would assume it would do?  I'm not familiar with the logic of how these tests are executed, I probably need to get them running to understand what's going wrong here.

> * What should `fumPrice()` be after doubling the size of the `ethPool` via `fund()`?

It should be unchanged.  Aside from fees, and aside from operations that nudge debt ratio above or below `MAX_DEBT_RATIO` (thereby setting or unsetting `minFumBuyPrice`), _none_ of `mint`/`burn`/`fund`/`defund` should move the FUM price at all: only ETH price changes should.  (See my [recent comment](https://github.com/alexroan/USM/pull/4/files#r483082729).)

jacob-eliosoff:
Once the tests are passing, should we merge this PR into master?  You guys have done a lot since master!

albertocuestacanada:
Yes, that's exactly the goal. Get the tests to pass and merge. Then we can add new features or refactor code in separate PRs.

contracts/BufferedToken.sol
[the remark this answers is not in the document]
jacob-eliosoff: OK, at least the conceptual distinction between "a pool of [some unspecified asset]" and "a subtype which is specifically a pool of ETH" makes sense to me now.  But, there are complexity pros & cons each way...  My instinct would be to only introduce that separation if we're ever actually trying to build a pool with a different asset.  Anyway yeah as @albertocuestacanada said let's carry on with the current structure and consider merging once we're back on master.  (Unless the split becomes a more immediate impediment to getting tests to pass etc.)

contracts/USM.sol
jacob-eliosoff: The easy ones are: `fund` decreases debt ratio, `defund` increases it.

The less easy ones: `mint` moves dr **closer to 100%** (so, increases it if <100%, decreases it if >100%), `burn` moves it **further from 100%** (the reverse).  This takes a bit of getting your head around but the main point is that (fees aside) **neither of them changes the size of the buffer.**

One way to think of it is, you can think of the ETH pool as made of two parts: the "covered" portion required to redeem all outstanding USM, and the "uncovered" buffer.  When dr > 100%, the buffer is negative (the system is underwater).  Two `mint` examples show the difference:

1. 20 ETH in pool = 10 ETH covered + 10 ETH buffer: dr = 50%.  `minting` 80 ETH means now 100 in pool = 90 covered + 10 buffer: dr = 90%.  So dr increased.
2. 20 in pool = 30 covered + -10 buffer: dr = 150%.  `minting` 80 means now 100 in pool = 110 covered + -10 buffer: dr = 110%.  So dr decreased.

And the `burn` examples are just the same thing played backwards:

3. 100 in pool = 90 covered + 10 buffer: dr = 90%.  `burning` 80 means now 20 in pool = 10 covered + 10 buffer: dr = 50%.  So dr decreased.
4. 100 in pool = 110 covered + -10 buffer: dr = 110%.  `burning` 80 means now 20 in pool = 30 covered + -10 buffer: dr = 150%.  So dr increased.

jacob-eliosoff: It is worth wrapping your heads around this because I anticipate you guys presenting about it at a conference sometime :)

contracts/USM.sol
```diff
@@ -1,117 +1,133 @@
...
-        uint ethAmount = _usmToEth(_usmAmount);
-        uint ethMinusFee = ethAmount.sub(mulFixed(ethAmount, BURN_FEE));
-        ethPool = ethPool.sub(ethMinusFee);
-        _burn(msg.sender, _usmAmount);
-        Address.sendValue(msg.sender, ethMinusFee);
+    function burn(uint _usmToBurn) external returns (uint) {
+        require(_usmToBurn >= MIN_BURN_AMOUNT, "Must burn at least 1 USM");
+        uint ethToSend = _burn(_usmToBurn);
+        require(debtRatio() <= WAD,
+            "Cannot burn with debt ratio below 100%");
```
jacob-eliosoff, line 74: "below 100%" -> "above 100%"

contracts/USM.sol
```diff
@@ -1,117 +1,136 @@
...
-     * @param _ethAmount The amount of ETH to convert.
-     * @return The amount of USM.
+     * @notice Funds the pool with ETH, minting FUM at its current price
      */
-    function _ethToUsm(uint _ethAmount) internal view returns (uint) {
-        require(_ethAmount > 0, "Eth Amount must be more than 0");
-        return mulFixed(_oraclePrice(), _ethAmount);
+    function _fund(address to, uint ethIn) internal {
+        uint _fumPrice = fumPrice();
+        uint fumOut = _fumPrice.wadMul(ethIn);
```
jacob-eliosoff, line 105: There's a bug here: should be `fumOut = ethIn.wadDiv(_fumPrice)`.  Eg, if the user passes in 3 ETH, and the current FUM price (in ETH terms) is 0.01 ETH, then the number of FUM created should be 3 / 0.01 = 300, not 0.01 * 3 = 0.03.

jacob-eliosoff:
The second problem with the `allows minting FUM` test is more subtle: the value of the ETH passed to `fund()`, and the value of the FUM returned, may not match!  This is because the buffer can have value even **before** the first `fund` operation.  Example:

1. A `mints` 10 ETH @ $200, gets back 2,000 USM
2. Price rises $200 -> $300: pool value increases $2,000 -> $3,000, buffer $0 -> $1,000
3. B `funds` 1 ETH @ $300: pool value $3,000 -> $3,300, buffer $1,000 -> $1,300
4. But now how many FUM should B get back?  The ETH they deposited was worth $300, but the buffer the FUM represents a claim on is worth $1,300!

The big picture here is this is only an issue for the very first `fund()` call, so it's not a huge deal.  The cleanest fix is probably to require that at least one `fund` op be done before the first `mint`, so that there's no "pre-existing buffer" like above.

I'm able to run the tests now so I'll aim to submit an actual code change in the next couple days.  Then we can finally close this PR!

jacob-eliosoff:
OK, the tests should be passing now.  My [two commits](https://github.com/alexroan/USM/commit/9774fcc7f6b6e2cd1a03baaa9585168196eaacdc) here were to USM.sol and 03_USM.test.js.

contracts/USM.sol
```diff
@@ -1,117 +1,138 @@
...
-     * @param _usmAmount The amount of USM to convert.
-     * @return The amount of ETH.
+     * @notice Defunds the pool by sending FUM out in exchange for equivalent ETH
+     * from the pool
      */
-    function _usmToEth(uint _usmAmount) internal view returns (uint) {
-        require(_usmAmount > 0, "USM Amount must be more than 0");
-        return divFixed(_usmAmount, _oraclePrice());
+    function defund(uint _fumAmount) external {
+        require(_fumAmount >= WAD, "Must defund at least 1 FUM");
```
jacob-eliosoff, line 118: I think we should just remove this line.  1 FUM could in principle ending up being pretty valuable!  And I don't see any reason not to let them redeem 0.000001 FUM if that's how they want to waste their gas.

contracts/BufferedToken.sol
```diff
@@ -0,0 +1,109 @@
...
+    using SafeMath for uint;
+    using WadMath for uint;
+
+    uint constant WAD = 10 ** 18;
+
+    uint public ethPool = 0;
```
jacob-eliosoff, line 21: Is the `ethPool` variable redundant with `address(this).balance`, ie, the actual amount of ETH currently stored by the contract?  Maybe we should just use the latter instead, ie, define a function `ethPool()` that just returns `address(this).balance`?

contracts/USM.sol
jacob-eliosoff: Easy so I pushed a fix.  (Always gotta remember to update the test too...)

contracts/BufferedToken.sol
albertocuestacanada: That will be the case once we merge `BuffereToken.sol` and `USM.sol` into one.

Right now, it can't be guaranteed that `eth` actually means ether, it could be any random ERC20. Which could be a useful feature to keep!

alexroan:
Tests passing locally for me. I'm happy to merge. If there are no objections I'll resolve the remaining conversations and go ahead with it.

contracts/BufferedToken.sol
jacob-eliosoff: I see.  Yeah, I don't see us using anything other than ETH (or WETH?) anytime soon but I see your point in terms of the design.

PR #6, opened 2020-08-30
Approximate power functions

albertocuestacanada:
Implementation of the approximate exponential functions from [the python source](https://github.com/jacob-eliosoff/usm-stablecoin/blob/751bbac83910fb5d1ae4ec91d740fd6b0dd9ed1f/usm_constproduct.py#L316).


albertocuestacanada:
To implement the half exponentials I imported the `DecimalMath.sol` contract I've been working on, which is strongly typed. Let's give it a spin.

All fixed point numbers have 27 digits, but that shouldn't be visible in USM.sol or other contracts. Functions should take a `UFixed memory` parameter when the number in question is a fixed point number.

`DecimalMath.sol` is a contract to cut on a lot of the boilerplate. `USM.sol` and others should inherit from it.

If we get tired of the boilerplate, we can always cut down to a weakly typed library, no big issue.

albertocuestacanada:
Trying to convert everything to use the strongly-typed `DecimalMath.sol` resulted in a nightmare. I'll downgrade to weakly-typed.

PR #7, opened 2020-08-30
Removed MINT_FEE and BURN_FEE - probably won't be using them

jacob-eliosoff:


albertocuestacanada (approved):
Compiles, and tests pass, good job.

PR #9, opened 2020-09-01
BufferedToken & USM - Split checks and transfers from business logic

albertocuestacanada:
I moved some code from `mint` and `burn` in `USM.sol` to `BufferedToken.sol` so that `BufferedToken` is the one managing the `ethPool` variable.

Also rewrote all the tests related to minting and burning.


alexroan (approved):
Love the refactor and the tests.

PR #10, opened 2020-09-01
Introduced a return statement to exit after 2 calls

alexroan:
Added a return statement so that `fund` doesn't call `_fund` 3 times in the scenario where both the following are true:
- `debtRatio` is greater than `MAX_DEBT_RATIO`
- `msg.value` is greater than `ethNeeded`


albertocuestacanada (approved):
Good catch

Issue #13, opened 2020-09-10
Run through a linter like solhint


jacob-eliosoff:
My first stab (like, never heard of solhint before 30 min ago) didn't turn up anything too horrifying:

![image](https://user-images.githubusercontent.com/8806519/92767920-96794d80-f365-11ea-9637-20fc224fc1f9.png)

albertocuestacanada:
Cool! Can you make this into a PR?

jacob-eliosoff:
Well, yeah the changes are easy, I'm just not 100% on most of them...  What is the correct visibility of a const like `WAD`, public?  Is it safe to rename Migration.sol's `last_completed_migration` to `lastCompletedMigration`?  Do we really need to shorten those USM error messages?  I could try but I'd be guessing a bit...

albertocuestacanada:
For now you can always use `public` for variables. Better to not touch `Migrations.sol` - good chance to check how to exclude lines or files for Solhint.

albertocuestacanada:
But if you do a PR, I can fix whichever errors you are not sure about.

jacob-eliosoff:
OK I will!

jacob-eliosoff:
See #14...  My first time creating a branch for a PR, hopefully don't bork anything

alexroan:
No Borks 😄

PR #14, opened 2020-09-10
Fixed up some solhint warnings.

jacob-eliosoff:


jacob-eliosoff:
See the output screenshot in [issue 13](https://github.com/alexroan/USM/issues/13).  I didn't touch Migrations.sol, I didn't shorten the error messages (they don't seem that long!), I didn't scrap the empty constructors (eg `constructor() public ERC20("Minimal Funding", "FUM") {}`), and I ignored the compiler version errors...

albertocuestacanada:
How did you install `solhint`? It's best to install it as a package using `package.json`, which would then be built using `yarn` and run using something like `npx solhint`.

That would allow all of us to use the same tools, and avoid having to set up everything individually for each one.

jacob-eliosoff:
Yeah I did install it that way.  I could include the changes to `package.json`, `package-lock.json` (?) and `.solhint.json` in the PR, they just seemed like a lot of cruft and I hope they're not too specific to my machine/setup...  But I guess I'll do so and you can judge for yourself.  The end result is I just run `npm run solhint`.  I was following https://medium.com/coinmonks/introduction-to-solidity-linting-and-formatting-e838c074791a (just via Google).

albertocuestacanada:
`package.json` and `.solhint.json` should be part of this PR. You can leave `package-lock.json` out.

albertocuestacanada (changes requested):
`package.json` and `.solhint.json` should be part of this PR. You can leave package-lock.json out.

jacob-eliosoff:
Added.

albertocuestacanada:
Happy with it.

 - Disabled the `compiler-version` rule, since we don't know what it does.
 - Disabled the `no-empty-blocks` rule, since there is no alternative.
 - `public` state variables are safe to use (`public` means "readable") and the compiler creates getters for them.

Use `yarn lint:sol` for solidity linting, and `yarn lint:js` for js linting (which auto-fixes).

albertocuestacanada:
Given @jacob-eliosoff absence, and low impact, I'm merging this.

PR #15, opened 2020-09-11
Merged BufferedToken into USM

### LINKED ISSUE #12: Merge BufferedToken and USM
The result should be simpler, let's check.

alexroan:
Fixes #12


contracts/USM.sol
```diff
@@ -131,4 +148,75 @@ contract USM is BufferedToken {
...
+    }
+
+    /**
+     * @notice Convert ETH amount to USM using the latest price of USM
+     * in ETH.
+     *
+     * @param _ethAmount The amount of ETH to convert.
+     * @return The amount of USM.
+     */
+    function _ethToUsm(uint _ethAmount) internal view returns (uint) {
```
albertocuestacanada, line 192: I would make this one public, it should be useful for users.

contracts/USM.sol
```diff
@@ -131,4 +148,75 @@ contract USM is BufferedToken {
...
+    }
+
+    /**
+     * @notice Convert USM amount to ETH using the latest price of USM
+     * in ETH.
+     *
+     * @param _usmAmount The amount of USM to convert.
+     * @return The amount of ETH.
+     */
+    function _usmToEth(uint _usmAmount) internal view returns (uint) {
```
albertocuestacanada, line 206: I would make this one public, it should be useful for users.

test/02_BufferedToken.test.js
```diff
@@ -1,157 +0,0 @@
-const { BN, expectRevert } = require('@openzeppelin/test-helpers')
```
albertocuestacanada, line 1: Instead of removing this file, create a mock contract that gives you access to the internal functions and keep the tests. If we refactor functionality then we can remove them one by one when they are not needed.

albertocuestacanada (changes requested):
I'm ok with the merge in principle, only I'm not sure we need to get rid of the BufferedToken tests.

contracts/USM.sol
```diff
@@ -131,4 +148,75 @@ contract USM is BufferedToken {
...
+
+    /** INTERNAL FUNCTIONS */
+
+    /**
+     * @notice Funds the pool with ETH, minting FUM at its current price
+     */
+    function _fund(address to, uint ethIn) internal {
+        uint _fumPrice = fumPrice();
+        uint fumOut = ethIn.wadDiv(_fumPrice);
+        ethPool = ethPool.add(ethIn);
```
albertocuestacanada, line 160: Have a go at removing `ethPool` and using `address(this).balance` instead.

contracts/USM.sol
```diff
@@ -131,4 +148,75 @@ contract USM is BufferedToken {
...
+        fum.mint(to, fumOut);
+        _setLatestFumPrice();
+    }
+
+    /**
+     * @notice Mint USM from Eth. Inheriting contract must handle the Ether transfers.
+     */
+    function _mint(uint ethAmount) internal returns (uint) {
```
albertocuestacanada, line 168: Once you remove `ethPool`, this function is reduced to `_mint(msg.sender, _ethToUsm(ethAmount));` and can be merged with `mint`.

contracts/USM.sol
```diff
@@ -131,4 +148,75 @@ contract USM is BufferedToken {
...
+        uint usmAmount = _ethToUsm(ethAmount);
+        ethPool = ethPool.add(ethAmount);
+        super._mint(msg.sender, usmAmount);
+        return usmAmount;
+    }
+
+    /**
+     * @notice Burn USM for Eth. Inheriting contract must handle the Ether transfers.
+     */
+    function _burn(uint usmAmount) internal returns (uint) {
```
albertocuestacanada, line 178: Once you remove `ethPool`, this function is reduced to `_burn(msg.sender, _usmToEth(usmAmount));` and can be merged with `burn`

albertocuestacanada (changes requested):
Actually, not a bad time to remove `ethPool`.

contracts/USM.sol
jacob-eliosoff: One thing to keep in mind here is that once we implement sliding prices, `ethToUsm()` becomes more complex...  Because suddenly the conversion from ETH to USM not only isn't linear any longer, but changes depending whether you're `minting`, `burning`, etc.  But we'll cross this bridge when we come to it.

test/02_BufferedToken.test.js
jacob-eliosoff: Yeah this seems worth at least trying...  To avoid scrapping useful tests.  Unless it becomes a big pain

contracts/USM.sol
```diff
@@ -1,38 +1,66 @@
...
+    /**
+     * @notice Calculate debt ratio of the current Eth pool amount and outstanding USM
+     * (the amount of USM in total supply).
+     *
+     * @return Debt ratio.
+     */
+    function debtRatio() public view returns (uint) {
+        if (ethPool == 0) {
+            return 0;
+        }
```
jacob-eliosoff, line 60: I'm not sure whether this is needed...  I _think_ once the contract is first funded with a `fund` op (and sliding prices are implemented), the pool is guaranteed to never shrink to 0?  But maybe including this check is safest...

contracts/USM.sol
```diff
@@ -131,4 +148,75 @@ contract USM is BufferedToken {
...
+     * @notice Convert USM amount to ETH using the latest price of USM
+     * in ETH.
+     *
+     * @param _usmAmount The amount of USM to convert.
+     * @return The amount of ETH.
+     */
+    function _usmToEth(uint _usmAmount) internal view returns (uint) {
+        if (_usmAmount == 0) {
+            return 0;
+        }
```
jacob-eliosoff, line 209: This check seems superfluous to me.  If they're silly enough to pass in 0 they can probably live with the same gas cost as a non-0 input.  Wdyt?

contracts/USM.sol
jacob-eliosoff: Yeah seems worth a shot...  This is the kind of refactoring our tests are for right!

contracts/USM.sol
jacob-eliosoff: Agreed (and generally with @albertocuestacanada's comments here)

jacob-eliosoff (approved):

contracts/USM.sol
```diff
@@ -72,7 +60,9 @@ contract USM is BufferedToken {
      */
     function burn(uint _usmToBurn) external returns (uint) {
         require(_usmToBurn >= MIN_BURN_AMOUNT, "1 USM minimum");
-        uint ethToSend = _burn(_usmToBurn);
+        // uint ethToSend = _burn(_usmToBurn);
+        uint ethToSend = usmToEth(_usmToBurn);
+        _burn(msg.sender, _usmToBurn);
         require(debtRatio() <= WAD, "Debt ratio too high");
```
alexroan, line 66: Removing `ethPool` in favour of the address balance has introduced an issue here.

Previously, our internal `_burn` made changes to the `ethPool` so that on this line, `debtRatio()` represents the future debt ratio after this transaction has completed.

Now, it won't because it uses the current balance, which will decrease a few lines later when the contract actually sends the ETH.

jacob-eliosoff: Ha!  Yeah I almost wrote a comment about this yesterday - removing `ethPool` means we lose control of exactly at what point in the logic it changes, which can introduce problems like you describe.  May have to hand-calculate debt ratio based on the upcoming ETH amount.  Or, in [`usm_constproduct.py`](https://github.com/jacob-eliosoff/usm-stablecoin/blob/master/usm_constproduct.py#L241) I dealt with this by giving `debt_ratio()` optional `eth` and `usm` params so you could pass in values other than the current actual ones, that's one possibility.

albertocuestacanada: I think that you could move the code from the internal _mint and _burn functions to the public ones, and put the debt ratio require afterwards, as a post-condition.

alexroan: Turns out this is resolved by moving the require statements to be post-conditions! :)

contracts/USM.sol
albertocuestacanada: Agreed, not needed.

contracts/USM.sol
albertocuestacanada: Agreed, not needed

albertocuestacanada (approved):

PR #17, opened 2020-09-16
feat: use weth

albertocuestacanada:
Using wrapped Eth instead of Eth itself, because dealing with ERC20 tokens is more convenient.

### DIFF
--- a/test/03_USM.test.js
+++ b/test/03_USM.test.js
@@ -1,8 +1,9 @@
 const { BN, expectRevert } = require('@openzeppelin/test-helpers')
 
-const USM = artifacts.require('./USM.sol')
-const FUM = artifacts.require('./FUM.sol')
-const TestOracle = artifacts.require('./TestOracle.sol')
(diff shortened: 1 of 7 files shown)


test/03_USM.test.js
```diff
@@ -23,69 +24,74 @@ contract('USM', (accounts) => {
     beforeEach(async () => {
       // Deploy contracts
       oracle = await TestOracle.new(price, shift, { from: deployer })
-      usm = await USM.new(oracle.address, { from: deployer })
+      weth = await WETH9.new({ from: deployer })
+      usm = await USM.new(oracle.address, weth.address, { from: deployer })
       fum = await FUM.at(await usm.fum())
+
+      // Set the user to obtain 10 Weth and approve USM to take them.
```
albertocuestacanada, line 31: The user needs to deposit Ether into the Weth contract to get Weth. Then he needs to approve USM as usual, for the ERC20 transfers to happen.

PR #16, opened 2020-09-11
Network/deploy rinkeby

albertocuestacanada:
To connect or deploy you will need a `.secret` file in your root with the 12 word mnemonic of your account, and an `.infuraKey` with the identifier of an infura.io account.

Get some rinkeby ether from https://faucet.rinkeby.io

Compile using `npx truffle compile`

Connect using `npx truffle console --network rinkeby`

Instantiate the contracts using:```
oracle = await TestOracle.at("0x80beEc0634819B555D6f4308068Bd83f7fd8ffDD")
usm = await USM.at("0xcc18a21186Ec39cf2Ffa5f19d393aC78b4ECc409")

image

A way of making your ether go far is to set the oracle price to a million times its normal value.

PR #18, opened 2020-09-16
Use real testnet oracles

albertocuestacanada:
Deployed in Rinkeby at 0x4E08A40F3ad891eC5700Af6D5E505E0A1A740B89

Instructions to connect in this PR: https://github.com/alexroan/USM/pull/16

Also verified in [Etherscan](https://rinkeby.etherscan.io/address/0xFc0165f11DA71D4d356CA8E766dd9AB7a0dB1a69#code), in case you prefer that as a frontend.


contracts/oracles/ChainlinkOracle.sol
```diff
@@ -5,7 +5,7 @@ import "@chainlink/contracts/src/v0.6/interfaces/AggregatorInterface.sol";
 
 contract ChainlinkOracle is IOracle{
 
-    address public oracle;
+    AggregatorInterface public oracle;
```
alexroan, line 8: `AggregatorInterface` is actually deprecated. For closer to real-time updates moving forward, we should use `AggregatorV3Interface`.

contracts/oracles/ChainlinkOracle.sol
```diff
@@ -14,11 +14,11 @@ contract ChainlinkOracle is IOracle{
...
      */
     constructor(address oracle_, uint shift) public {
-        oracle = oracle_;
+        oracle = AggregatorInterface(oracle_);
         decimalShift = shift;
     }
 
     function latestPrice() external override view returns (uint) {
-        return uint(AggregatorInterface(oracle).latestAnswer());
+        return uint(oracle.latestAnswer());
```
alexroan, line 22: `latestAnswer` from `AggregatorInterface` is actually deprecated. For closer to real-time updates moving forward, we should use `latestRoundData` from `AggregatorV3Interface`.

alexroan (changes requested):
We should update to the latest price feed interface, which is `AggregatorV3Interface`. This interface and the function `latestAnswer` has been recently deprecated.

alexroan (approved):

PR #20, opened 2020-09-19
feat: use wrapped ether

albertocuestacanada:
Just that, better description to come.

### DIFF
--- a/contracts/external/WETH9.sol
+++ b/contracts/external/WETH9.sol
@@ -1,0 +1,78 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (C) 2015, 2016, 2017 Dapphub
+
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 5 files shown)


contracts/external/WETH9.sol
```diff
@@ -0,0 +1,78 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
alexroan, line 1: Am I right in thinking this is purely for testing purposes?

If so could we use ERC20 from Openzeppelin instead of including this?

alexroan (commented):
Just one comment on the WETH.sol contract. If all is well, I'm happy for you to merge!

contracts/external/WETH9.sol
albertocuestacanada: Yes, this is solely for local testing purposes. If you check the migrations scripts, on those we connect to the existing deployments. We will never deploy `WETH9`. 

Now that `WETH9` extends `ERC20` with `deposit` and `withdraw` functions. 

The `WETH9` included is the same that is deployed in mainnet (and holds $2B). To make sure we test against the deployed bytecode, we must use exactly this contract.

PR #21, opened 2020-09-19
Feat: eth proxy - Fixes #11

### LINKED ISSUE #11: Refactor to use Weth
Using ETH is a pain, let's consider using Weth instead.

albertocuestacanada:
Implementation of an Eth <> Weth proxy for USM using Delegable.sol. This one I'll need to explain in depth.

### DIFF
--- a/contracts/EthProxy.sol
+++ b/contracts/EthProxy.sol
@@ -11,7 +11,9 @@ contract EthProxy {
     IUSM public usm;
     IWETH9 public weth;
 
-    constructor(address usm_, address weth_) public {
+    constructor(address usm_, address weth_)
+        public
+    {
         usm = IUSM(usm_); 
         weth = IWETH9(weth_);
 
@@ -24,9 +26,10 @@ contract EthProxy {
     /// @dev Users use `mint` in EthProxy to post ETH to USM (amount = msg.value), which will be converted to Weth here.
     /// @param to Receiver of the minted USM
     function mint(address to)
-        external payable {
+        external payable returns (uint)
+    {
         weth.deposit{ value: msg.value }();
-        usm.mint(address(this), to, msg.value);
+        return usm.mint(address(this), to, msg.value);
     }
 
     /// @dev Users wishing to withdraw their Weth as ETH from USM should use this function.
@@ -34,10 +37,12 @@ contract EthProxy {
     /// @param to Receiver of the obtained Eth
     /// @param usmToBurn Amount of USM to burn.
     function burn(address payable to, uint usmToBurn)
-        external {
+        external returns (uint)
+    {
         uint wethToWithdraw = usm.burn(msg.sender, address(this), usmToBurn);
         weth.withdraw(wethToWithdraw);
         to.sendValue(wethToWithdraw);
+        return wethToWithdraw;
     }
 
     /**
@@ -45,10 +50,10 @@ contract EthProxy {
      * @param to Receiver of the obtained FUM
      */
     function fund(address to)
-        external payable
+        external payable returns (uint)
     {
         weth.deposit{ value: msg.value }();
-        usm.fund(address(this), to, msg.value);
+        return usm.fund(address(this), to, msg.value);
     }
 
     /**
@@ -57,10 +62,11 @@ contract EthProxy {
      * @param fumToBurn Amount of FUM to burn.
      */
     function defund(address payable to, uint fumToBurn)
-        external
+        external returns (uint)
     {
         uint wethToWithdraw = usm.defund(msg.sender, address(this), fumToBurn);
         weth.withdraw(wethToWithdraw);
         to.sendValue(wethToWithdraw);
+        return wethToWithdraw;
     }
 }
--- a/contracts/EthProxy.sol
+++ b/contracts/EthProxy.sol
@@ -12,6 +12,8 @@ contract EthProxy {
     constructor(address usm_, address weth_) public {
         usm = IUSM(usm_); 
         weth = IWETH9(weth_);
+
+        weth.approve(address(usm), uint(-1));
     }
 
     /// @dev The WETH9 contract will send ether to EthProxy on `weth.withdraw` using this function.
@@ -29,9 +31,9 @@ contract EthProxy {
     /// Users must have called `controller.addDelegate(ethProxy.address)` to authorize EthProxy to act in their behalf.
     /// @param to Receiver of the obtained Eth
     /// @param usmToBurn Amount of USM to burn.
-    function burn(address payable to, uint256 usmToBurn)
+    function burn(address payable to, uint usmToBurn)
         external {
-        uint256 wethToWithdraw = usm.burn(msg.sender, address(this), usmToBurn);
+        uint wethToWithdraw = usm.burn(msg.sender, address(this), usmToBurn);
         weth.withdraw(wethToWithdraw);
         to.transfer(wethToWithdraw);
     }
@@ -55,7 +57,7 @@ contract EthProxy {
     function defund(address payable to, uint fumToBurn)
         external
     {
-        uint256 wethToWithdraw = usm.defund(msg.sender, address(this), fumToBurn);
+        uint wethToWithdraw = usm.defund(msg.sender, address(this), fumToBurn);
         weth.withdraw(wethToWithdraw);
         to.transfer(wethToWithdraw);
     }
--- a/contracts/EthProxy.sol
+++ b/contracts/EthProxy.sol
@@ -1,11 +1,13 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 pragma solidity ^0.6.7;
 
+import "@openzeppelin/contracts/utils/Address.sol";
 import "./IUSM.sol";
 import "./external/IWETH9.sol";
 
 
 contract EthProxy {
+    using Address for address payable;
     IUSM public usm;
     IWETH9 public weth;
 
@@ -35,7 +37,7 @@ contract EthProxy {
         external {
         uint wethToWithdraw = usm.burn(msg.sender, address(this), usmToBurn);
         weth.withdraw(wethToWithdraw);
-        to.transfer(wethToWithdraw);
+        to.sendValue(wethToWithdraw);
     }
 
     /**
@@ -59,6 +61,6 @@ contract EthProxy {
     {
         uint wethToWithdraw = usm.defund(msg.sender, address(this), fumToBurn);
         weth.withdraw(wethToWithdraw);
-        to.transfer(wethToWithdraw);
+        to.sendValue(wethToWithdraw);
     }
 }
--- a/contracts/EthProxy.sol
+++ b/contracts/EthProxy.sol
@@ -1,0 +1,62 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.6.7;
+
+import "./IUSM.sol";
+import "./external/IWETH9.sol";
+
+
+contract EthProxy {
+    IUSM public usm;
+    IWETH9 public weth;
+
+    constructor(address usm_, address weth_) public {
+        usm = IUSM(usm_); 
+        weth = IWETH9(weth_);
+    }
+
+    /// @dev The WETH9 contract will send ether to EthProxy on `weth.withdraw` using this function.
+    receive() external payable { }
+
+    /// @dev Users use `mint` in EthProxy to post ETH to USM (amount = msg.value), which will be converted to Weth here.
+    /// @param to Receiver of the minted USM
+    function mint(address to)
+        external payable {
+        weth.deposit{ value: msg.value }();
+        usm.mint(address(this), to, msg.value);
+    }
+
+    /// @dev Users wishing to withdraw their Weth as ETH from USM should use this function.
+    /// Users must have called `controller.addDelegate(ethProxy.address)` to authorize EthProxy to act in their behalf.
+    /// @param to Receiver of the obtained Eth
+    /// @param usmToBurn Amount of USM to burn.
+    function burn(address payable to, uint256 usmToBurn)
+        external {
+        uint256 wethToWithdraw = usm.burn(msg.sender, address(this), usmToBurn);
+        weth.withdraw(wethToWithdraw);
+        to.transfer(wethToWithdraw);
+    }
+
+    /**
+     * @notice Funds the pool with ETH, converted to WETH
+     * @param to Receiver of the obtained FUM
+     */
+    function fund(address to)
+        external payable
+    {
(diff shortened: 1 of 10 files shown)


albertocuestacanada:
Tests complete, let's discuss.

contracts/USM.sol
```diff
@@ -12,7 +14,7 @@ import "./oracles/IOracle.sol";
  * @author Alex Roan (@alexroan)
  * @notice Concept by Jacob Eliosoff (@jacob-eliosoff).
  */
-contract USM is ERC20 {
+contract USM is ERC20, Delegable {
```
jacob-eliosoff, line 17: Do you not need `USM is IUSM`?

contracts/EthProxy.sol
```diff
@@ -0,0 +1,66 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.6.7;
+
+import "@openzeppelin/contracts/utils/Address.sol";
+import "./IUSM.sol";
+import "./external/IWETH9.sol";
+
+
+contract EthProxy {
```
jacob-eliosoff, line 9: And how about here?  `contract EthProxy is IUSM`?

contracts/USM.sol
```diff
@@ -74,31 +84,41 @@ contract USM is ERC20 {
...
         require(ethIn > MIN_ETH_AMOUNT, "0.001 ETH minimum"); // TODO: Needed?
         if(debtRatio() > MAX_DEBT_RATIO){
             // calculate the ETH needed to bring debt ratio to suitable levels
             uint ethNeeded = usmToEth(totalSupply()).wadDiv(MAX_DEBT_RATIO).sub(ethPool()).add(1); //+ 1 to tip it over the edge
+            uint fumOut;
             if (ethIn >= ethNeeded) { // Split into two fundings at different prices
-                _fund(msg.sender, ethNeeded);
-                _fund(msg.sender, ethIn.sub(ethNeeded));
-                return;
+                fumOut = _fund(to, ethNeeded);
```
jacob-eliosoff, line 99: Probably simplify 97-99 to:

```
            if (ethIn >= ethNeeded) { // Split into two fundings at different prices
                uint fumOut = _fund(to, ethNeeded);
```

jacob-eliosoff (commented):
Not the most thorough review in the world but a couple comments...  In general the idea makes sense to me

contracts/USM.sol
albertocuestacanada: Not really, and `interface` is a declaration of the methods contained in a contract, but the contract itself doesn't need to inherit the interface for it to be valid. Actually, the contract inheriting the interface makes the code more unreadable (by demanding `override` in each function).

However, to follow general practice, and maybe to get some compiler warnings if something goes wrong, I did `USM is IUSM`.

contracts/EthProxy.sol
albertocuestacanada: Not here. `EthProxy` is a standalone contract, that contains a `IUSM` link inside. Same as `USM` doesn't inherit from `WETH9`, it takes its address as a constructor parameter and then calls it when needed.

It's important to understand that `EthProxy` is completely independent from `USM` and has no special access beyond what each individual user allows.

contracts/USM.sol
albertocuestacanada: Simplified to a one-liner, hopefully readable enough.

contracts/EthProxy.sol
jacob-eliosoff: Interesting.  I was operating on a pre-smart-contract intuition where one might define an interface `IUSM` and then have two implementations: one that takes WETH, and another that takes ETH (and passes the input to a wrapped instance of the one that takes WETH).  And then other functions/code could operate on variables of type `IUSM`, without caring which of the two concrete types the var's value was an instance of...

...But after reading a bit more, this seems ambitious for Solidity's/the EVM's type system, and probably overkill for our purposes anyway.  Suggestion retracted!

contracts/EthProxy.sol
```diff
@@ -0,0 +1,66 @@
...
+    /// @dev The WETH9 contract will send ether to EthProxy on `weth.withdraw` using this function.
+    receive() external payable { }
+
+    /// @dev Users use `mint` in EthProxy to post ETH to USM (amount = msg.value), which will be converted to Weth here.
+    /// @param to Receiver of the minted USM
+    function mint(address to)
+        external payable {
+        weth.deposit{ value: msg.value }();
+        usm.mint(address(this), to, msg.value);
+    }
```
jacob-eliosoff, line 30: If it's useful for `USM.mint()` to return the amount minted, might it not also be useful for `EthProxy.mint()` to return the same (ie, `usm.mint()`)?  Seems like regardless of whether they both formally implement the same interface, we might as well give them the same API.

I also wonder if names like `WethUsm` (or `WethUSM` or something, dunno) and `EthUsm` might be clearer than `USM` and `EthProxy`.

contracts/EthProxy.sol
```diff
@@ -0,0 +1,66 @@
...
+    /// @dev Users wishing to withdraw their Weth as ETH from USM should use this function.
+    /// Users must have called `controller.addDelegate(ethProxy.address)` to authorize EthProxy to act in their behalf.
+    /// @param to Receiver of the obtained Eth
+    /// @param usmToBurn Amount of USM to burn.
+    function burn(address payable to, uint usmToBurn)
+        external {
+        uint wethToWithdraw = usm.burn(msg.sender, address(this), usmToBurn);
+        weth.withdraw(wethToWithdraw);
+        to.sendValue(wethToWithdraw);
+    }
```
jacob-eliosoff, line 41: Same for consistency of the return value here, and for the other fns...

contracts/USM.sol
```diff
@@ -46,12 +45,16 @@ contract USM is ERC20 {
...
      */
-    function mint(uint256 ethIn) external returns (uint) {
-        require(eth.transferFrom(msg.sender, address(this), ethIn), "Eth transfer fail");
+    function mint(address from, address to, uint ethIn)
+        external override
+        onlyHolderOrDelegate(from, "Only holder or delegate")
+        returns (uint)
+    {
+        require(eth.transferFrom(from, address(this), ethIn), "Eth transfer fail");
         require(ethIn > MIN_ETH_AMOUNT, "0.001 ETH minimum");
```
jacob-eliosoff, line 54: As with some of these checks before - I'm not strongly against it per se - but what's the benefit?  Unless there's a clear one I'd be inclined to scrap random checks like this.  (Especially since who knows, in the future 0.001 ETH might be like a million dollars)

contracts/EthProxy.sol
```diff
@@ -0,0 +1,66 @@
...
+
+    constructor(address usm_, address weth_) public {
+        usm = IUSM(usm_); 
+        weth = IWETH9(weth_);
+
+        weth.approve(address(usm), uint(-1));
+    }
+
+    /// @dev The WETH9 contract will send ether to EthProxy on `weth.withdraw` using this function.
+    receive() external payable { }
```
jacob-eliosoff, line 22: Can you explain in like one line (just to me, not nec in the code) how this empty-body fn does anything?

contracts/EthProxy.sol
```diff
@@ -0,0 +1,66 @@
...
+
+    /// @dev Users use `mint` in EthProxy to post ETH to USM (amount = msg.value), which will be converted to Weth here.
+    /// @param to Receiver of the minted USM
+    function mint(address to)
+        external payable {
+        weth.deposit{ value: msg.value }();
+        usm.mint(address(this), to, msg.value);
```
jacob-eliosoff, line 29: Oy vey, the more closely I read this class the more confused I get!  😂  Probably just because this is all so new to me.  But: above we call `weth.approve(address(usm))`, and here we call `usm.mint(address(this), ...)`...  I'm trying to understand why that doesn't read `usm.mint(address(usm), ...)`.  Does `usm.mint()` end up transferring the WETH from `this`, or from `usm`?  Or does some magic make them the same.

test/03_USM.test.js
```diff
@@ -85,54 +85,60 @@ contract('USM', (accounts) => {
...
 
           it('allows burning FUM', async () => {
-            const fumBuyPrice = (await usm.fumPrice(sides.BUY)).toString()
-            const fumSellPrice = (await usm.fumPrice(sides.SELL)).toString()
+            const fumBuyPrice = (await usm.fumPrice(sides.BUY))
+            const fumSellPrice = (await usm.fumPrice(sides.SELL))
 
-            const fumBalance = (await fum.balanceOf(user)).toString()
+            const fumBalance = (await fum.balanceOf(user1)).toString()
             const targetFumBalance = oneEth.mul(priceWAD).div(WAD) // see "allows minting FUM" above
```
jacob-eliosoff, line 101: Maybe for consistency/clarity we should just never do `priceOrWhatev = x.toString(); x.should.equal(y.toString())`, and always do `priceOrWhatev = x; x.toString().should.equal(y.toString())`?

jacob-eliosoff: (Don't feel super strongly about this)

contracts/EthProxy.sol
albertocuestacanada: Fixed the return values.

albertocuestacanada: `EthProxy` will actually evolve towards just `Proxy`. In `USM.sol` we will have the core functions that allow USM to implement its business logic, in `Proxy.sol` all the functions that make it more convenient for users.

contracts/EthProxy.sol
albertocuestacanada: Fixed

contracts/EthProxy.sol
albertocuestacanada: Note that it is not `function receive()` but `receive()`.

Until solidity 0.6 the fallback function implemented the logic to be executed when receiving ether. In solidity 0.6 and later the logic to execute when receiving ether must be included in the special `receive()` function.

If the body of `receive()` is empty, it just does the standard fare, updating balances and whatever.

`receive()` needs to exist in `EthProxy.sol`, so that it can receive ether from WETH9 on executing `weth.withdraw()`

contracts/EthProxy.sol
albertocuestacanada: `EthProxy` acts as a buffer here.

The user sends ether to `EthProxy` using `proxy.mint(receiver, { value: amount })`.
`EthProxy` converts the ether to weth, in his own account, using `weth.deposit({ value: amount })`
Now that `EthProxy` has weth, it uses it to mint USM in the chosen USM account: `usm.mint(from, to, amount)` -> `usm.mint(address(this), to, amount)` -> `usm.mint(proxy.address, to, amount)`

You might be confused by the existence of `from` and `to` parameters in all functions, that allows to use any function as a means of payment.

The most common use case would be `proxy.mint(user1, { from: user1, value: amount})`

test/03_USM.test.js
albertocuestacanada: Fixed, values should be BigNumber (or BN) for the whole test, and converted to string only for and at the assertion.

jacob-eliosoff: Nice!  Thank you.

contracts/EthProxy.sol
jacob-eliosoff: Yeah I understand _most_ of that.  My specific confusion is, does the `weth.deposit()` call result in the WETH being added to `address(weth)`, or to `address(this)`?  (Since the `usm.mint(address(this), ...)` call is spending it from `this`.)  Or does some magic mean adding WETH to `weth` is identical to adding it to `this`.

contracts/EthProxy.sol
jacob-eliosoff: Ah yeah OK I did read about `receive()` vs `fallback()` at some point but I forgot about [this part](https://solidity.readthedocs.io/en/latest/contracts.html#receive-ether-function): "If neither a receive Ether nor a payable fallback function is present, the contract cannot receive Ether through regular transactions and throws an exception."  Got it now, thanks.

contracts/EthProxy.sol
jacob-eliosoff: Well I suppose it depends whether we want names that reflect how it works internally, or what service it provides to users.  From a user's pov, they're just two similar contracts, one takes ETH the other WETH.  Anyway my larger point was that EthProxy (and even more so Proxy) is a very generic non-USM-specific name: in other languages it would raise some risk of name conflicts with other libraries etc, but maybe in Solidity that's not a concern.

contracts/EthProxy.sol
jacob-eliosoff: Maybe my confusion is that `weth.deposit()` doesn't mean "Add WETH to the `weth` address", it means "Use `weth` to add WETH to this contract's address"?

contracts/EthProxy.sol
albertocuestacanada: The name code is not a concern. A conflict would happen only if we would have another contract with the same name. It doesn't matter if someone else chooses the same name for their contract.

On function names, I can append `FromEth` and `ToEth` to the methods in `Proxy`, if that helps. However, I expect that someone calling methods manually should notice that one contract is `USM` and the other `Proxy`, and be able to read the documentation to know what each function does.

jacob-eliosoff: No the function names are fine.  They _should_ be the same!

contracts/EthProxy.sol
albertocuestacanada: `weth.deposit()` means "Hey, weth contract, take my ETH and give me Weth". `Proxy` takes the ETH from the user, calls `weth.deposit` which credits Weth to the `Proxy` address, and then `Proxy` calls `usm.mint(address(this), to, amount)` to mint USM with its Weth, into the `to` account that the user specified.

So, for an instant, it's `Proxy` that has Weth, which then uses immediately to mint USM in favour of `to`.

contracts/EthProxy.sol
albertocuestacanada: Should I rename `EthProxy` as `Proxy` already?

contracts/USM.sol
albertocuestacanada: I'm always in favor of removing code :+1:

contracts/EthProxy.sol
jacob-eliosoff: I defer completely to you (guys) on this one!

contracts/EthProxy.sol
jacob-eliosoff: Yup, OK, that's what my last comment referred to.  I had read `weth.deposit()` as "Deposit into `weth`", but in fact it's "Deposit into me."  So it all makes sense to me now, thanks.

PR #28, opened 2020-09-23
feat: enabled erc20permit - Fixes #22

### LINKED ISSUE #22: USM and FUM inherit from ERC20Permit

albertocuestacanada:
Added allowance changes through `permit`.

@alexroan, you can use the `node_modules/erc20permit/utils/signatures.ts` script to integrate signatures in the UI. It is quite some work to get it right, ask for help whenever needed. Also, you can use the `eth-permit` npm package from David Mihal instead if you want.

### DIFF
--- a/contracts/FUM.sol
+++ b/contracts/FUM.sol
@@ -1,7 +1,7 @@
 pragma solidity ^0.6.7;
 
-import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 import "@openzeppelin/contracts/access/Ownable.sol";
+import "erc20permit/contracts/ERC20Permit.sol";
 
 /**
  * @title FUM Token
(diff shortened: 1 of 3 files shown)


alexroan:
So, is the purpose of using ERC20Permit to enable GAS-less transactions for our users?

alexroan (commented):
Changes look completely fine. Can you clarify the benefits of using ERC20Permit instead of ERC20?

Is the primary reason for it to enable GAS-less transactions for users?

albertocuestacanada:
The main benefit that I see is in terms of UX.

Currently, with any ERC20, first you have to send an `approve` transaction, and then you do whatever you actually intended to do, like `fund`. 
 - click button - accept popup - 30 second wait - click button - accept popup - 30 second wait - done!

Using off-chain signatures you don't really save much gas, if any, but you can bundle both transactions into one.
 - click button - accept popup 1 - accept popup 2 - 30 second wait - done!

We have 2 ERC20 that can be approved via off-chain signature, plus the permissions that the user needs to give the proxy to act in his behalf. Instead of each one of those being a separate transaction, they can be just popups.

alexroan:
OK, that makes a lot of sense. I'll probably ping you again regarding this when I'm working on the UI. 👍

PR #29, opened 2020-09-23
fix: added spdx licenses where needed

albertocuestacanada:
Just a small fix to add the licenses where they were missing.


jacob-eliosoff:
Looks fine to me...  Is this worth adding to the tests too?  I have no idea

alexroan (approved):

PR #34, opened 2020-09-24
Feat/limit orders - Fixes #24

### LINKED ISSUE #24: fund limit buy orders
fund limit buy orders (aka: specify a max FUM price when doing a fund op, so that the ETH/USM is stored for later execution if the current FUM price is too high)

albertocuestacanada:
Decided to extend the limit feature to Eth orders as well as Weth orders.

Since `Proxy` acts as an intermediary holder for the Eth / Weth conversions, the code has to be duplicated for Eth and for Weth (`burn` and `burnForEth` do different calls to `usm.burn`)

The renaming of `EthProxy` as `Proxy` messed up with the diff, sorry :(

### DIFF
--- a/contracts/Proxy.sol
+++ b/contracts/Proxy.sol
@@ -6,7 +6,7 @@ import "./IUSM.sol";
 import "./external/IWETH9.sol";
 
 
-contract EthProxy {
+contract Proxy {
     using Address for address payable;
     IUSM public usm;
     IWETH9 public weth;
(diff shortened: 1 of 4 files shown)


alexroan (approved):
Looks good to me, as tidy as it could be!

Entertained the idea of using overflow functions, so instead of having:

* `fund()`
* `fundWithEth()`

We'd have:

* `fund()`
* `fund()`

but with different parameters depending on whether the user is sending WEth or Eth.

Problem is we have a `burn` function with the same inputs as `burnWithEth` (address, uint, uint) so the EVM wouldn't be able to tell the difference.

Happy for this to go in! :) Thanks @albertocuestacanada

jacob-eliosoff:
There's a case for renaming our `mint()`/`burn()`-type functions to clearer things like `createFumFromEth()` or `redeemUsmForEth()`:
1. I just tripped over a [bug](https://github.com/alexroan/USM/commit/ec8b3935cfb770d1ae617a6e250574f2d58ebefb) at least partly due to confusion over whether `burn()` refers to burning USM or FUM.
2. More explicit names would make it cleaner to add a function like `createFumFromUsm()` (which I think we do want).
3. We wouldn't need to rename anything in the UI (webpage etc), just in the code.
4. When I named them `mint()` and `burn()` in my original post, I didn't appreciate that those were already reserved terms in Solidity/ERC20s!  This has created some confusion for me before, when eg our own `burn()` fn calls the internal `_burn()` fn.

But, I don't feel strongly about this, we could keep it in the back of our minds for now.

PR #35, opened 2020-09-27
Major changes: sliding prices, mint/burn/fund/defund adjustments, ...

jacob-eliosoff:
Still a little rough - I may give it another look-through and see if there are places I can add comments or otherwise improve.  But hey the tests pass.  Please nitpick heavily and let me know any questions or suggestions!


contracts/USM.sol
```diff
@@ -148,21 +179,149 @@ contract USM is IUSM, ERC20Permit, Delegable {
...
+     * @return The amount of ETH, and the factor by which the ethPool shrank.
+     */
+    function ethFromBurn(uint usmIn) public override view returns (uint, uint) {
+        // Burn at a sliding-up price:
+        uint initialUsmPrice = usmPrice(Side.Sell);
+        uint _ethPool = ethPool();
+        // Math: this is an integral - sum of all USM burned at a sliding price.  Follows the same mathematical invariant as
+        // above: if pool_eth *= k, eth_price *= 1/k**2.
+        uint ethOut = WAD.wadDiv(WAD.wadDiv(_ethPool) + WAD.wadDiv(usmIn.wadMul(initialUsmPrice)));
+        uint ethPoolShrinkFactor = (_ethPool - ethOut).wadDiv(_ethPool);
```
albertocuestacanada, line 284: Shouldn't we check that `ethPool` >= `ethOut`?

contracts/USM.sol
```diff
@@ -148,21 +179,149 @@ contract USM is IUSM, ERC20Permit, Delegable {
...
+        uint ethPoolGrowthFactor;
+        uint fumOut;
+        if (_ethPool == 0 ) {
+            // This is our first ETH added, so an "ethPoolGrowthFactor" makes no sense - skip sliding-prices for this first call:
+            ethPoolGrowthFactor = type(uint).max;
+            fumOut = ethIn.wadDiv(initialFumPrice);
+        } else {
+            ethPoolGrowthFactor = (_ethPool + ethIn).wadDiv(_ethPool);
+            // Math: see closely analogous comment in mint_usm() above.
+            fumOut = _ethPool.wadDiv(initialFumPrice).wadMul(WAD - WAD.wadDiv(ethPoolGrowthFactor));
```
albertocuestacanada, line 307: `ethPoolGrowthFactor` can't be less than 1, so no need for `SafeMath.sol` here either.

test/03_USM.test.js
```diff
@@ -29,127 +43,280 @@ contract('USM', (accounts) => {
...
+        const fumBuyPrice2 = (await usm.fumPrice(sides.BUY))
+        fumBuyPrice2.toString().should.equal(fumBuyPrice.toString()) // First fund() (w/o sliding prices) doesn't change FUM price
+        const fumSellPrice2 = (await usm.fumPrice(sides.SELL))
+        fumSellPrice2.toString().should.equal(fumSellPrice.toString())
 
-        const newFumBuyPrice = (await usm.fumPrice(sides.BUY))
-        newFumBuyPrice.toString().should.equal(fumBuyPrice.toString()) // Funding doesn't change the fum price
-        const newFumSellPrice = (await usm.fumPrice(sides.SELL))
-        newFumSellPrice.toString().should.equal(fumSellPrice.toString())
+        await usm.fund(user1, user2, oneEth, { from: user1 })
```
albertocuestacanada, line 94: Please split this second half of the test into a test of its own, inside the "with existing FUM supply" `describe`.

albertocuestacanada, line 94: Here, instead of funding `oneEth`, retrieve the `ethPool` value and fund by that. Then the test will be done on doubling the `ethPool`, whatever the initial `fund` was.

Let's try to parameterize the tests.

test/03_USM.test.js
```diff
@@ -29,127 +43,280 @@ contract('USM', (accounts) => {
...
+        const ethPool3 = (await weth.balanceOf(usm.address))
+        ethPool3.toString().should.equal(oneEth.mul(TWO).toString())
+
+        const fumBalance3 = (await fum.balanceOf(user2))
+        // Calls to fund() after the first start using sliding prices.  In particular, following the math in USM.fumFromFund(),
+        // with the second call here 2xing the pool (1 ETH -> 2) and the initial FUM price of $1 = 0.004 ETH, the FUM added should
+        // be 1 (current ETH pool) / 0.004 * (1 - 1 / 2) = 125, ie, the user's FUM balance should increase by a factor of 3/2:
+        fumBalance3.toString().should.equal(wadMul(oneEth, priceWAD).mul(THREE).div(TWO).toString())
```
albertocuestacanada, line 102: As before, the expected FUM increase should be calculated in the test, instead of hardcoded to `3/2`.

Also, please store the FUM supply in a variable, instead of using `wadMul(oneEth, priceWAD)`

test/03_USM.test.js
```diff
@@ -29,127 +43,280 @@ contract('USM', (accounts) => {
...
+        const block0 = (await web3.eth.getBlockNumber())
+        const t0 = (await web3.eth.getBlock(block0)).timestamp
+        const timeDelay = 3 * 60
+        await timeMachine.advanceTimeAndBlock(timeDelay)
+        const block1 = (await web3.eth.getBlockNumber())
+        const t1 = (await web3.eth.getBlock(block1)).timestamp
+        t1.toString().should.equal((t0 + timeDelay).toString())
+        const fundDefundAdj4 = (await usm.fundDefundAdjustment())
+        const decayFactor4 = WAD.div(EIGHT)
+        const targetFundDefundAdj4 = WAD.add(wadMul(fundDefundAdj3, decayFactor4)).sub(decayFactor4)
```
albertocuestacanada, line 133: I think it might be useful to code this in a helper function, same as you did with `wadMul`.

test/03_USM.test.js
```diff
@@ -29,127 +43,280 @@ contract('USM', (accounts) => {
...
+        fundDefundAdj4.divRound(TEN).toString().should.equal(targetFundDefundAdj4.divRound(TEN).toString())
       })
 
-      describe('with existing FUM supply', () => {
+      describe("with existing FUM supply", () => {
         beforeEach(async () => {
           await usm.fund(user1, user1, oneEth, { from: user1 })
```
albertocuestacanada, line 139: Should we use `ethPool()` instead of `oneEth`, to parameterize?

test/03_USM.test.js
```diff
@@ -29,127 +43,280 @@ contract('USM', (accounts) => {
...
         })
 
-        it('allows minting USM', async () => {
-          const fumBuyPrice = (await usm.fumPrice(sides.BUY))
-          const fumSellPrice = (await usm.fumPrice(sides.SELL))
-
+        it("allows minting USM", async () => {
           await usm.mint(user1, user2, oneEth, { from: user1 })
```
albertocuestacanada, line 143: Should we use `ethPool()` instead of `oneEth`, to parameterize?

test/03_USM.test.js
```diff
@@ -29,127 +43,280 @@ contract('USM', (accounts) => {
...
-          const newFumSellPrice = (await usm.fumPrice(sides.SELL))
-          newFumSellPrice.toString().should.equal(fumSellPrice.toString())
+          const ethPool2 = (await weth.balanceOf(usm.address))
+          ethPool2.toString().should.equal(oneEth.mul(TWO).toString())
+
+          const usmBalance2 = (await usm.balanceOf(user2))
+          // Verify that sliding prices are working: this mirrors the sliding FUM prices above - see the math in USM.usmFromMint(),
+          // with mint() here 2xing the pool (1 ETH from fund() earlier -> 2) and the initial USM price of $1 = 0.004 ETH, the USM
+          // added should be 1 (current ETH pool) / 0.004 * (1 - 1 / 2) = 125 (ie, the ETH price $250 / 2):
+          usmBalance2.toString().should.equal(wadMul(oneEth, priceWAD).div(TWO).toString())
```
albertocuestacanada, line 151: Can we code the target value into a formula?

test/03_USM.test.js
```diff
@@ -29,127 +43,280 @@ contract('USM', (accounts) => {
...
+          const block0 = (await web3.eth.getBlockNumber())
+          const t0 = (await web3.eth.getBlock(block0)).timestamp
+          const timeDelay = 3 * 60
+          await timeMachine.advanceTimeAndBlock(timeDelay)
+          const block1 = (await web3.eth.getBlockNumber())
+          const t1 = (await web3.eth.getBlock(block1)).timestamp
+          t1.toString().should.equal((t0 + timeDelay).toString())
+          const mintBurnAdj3 = (await usm.mintBurnAdjustment())
+          const decayFactor3 = WAD.div(EIGHT)
+          const targetMintBurnAdj3 = WAD.add(wadMul(mintBurnAdj2, decayFactor3)).sub(decayFactor3)
```
albertocuestacanada, line 181: Should we extract this into a helper function?

test/03_USM.test.js
```diff
@@ -29,127 +43,280 @@ contract('USM', (accounts) => {
...
 
-          it('allows burning FUM', async () => {
-            const fumBuyPrice = (await usm.fumPrice(sides.BUY))
+          it("allows burning FUM", async () => {
+            const ethPool = (await usm.ethPool())
             const fumSellPrice = (await usm.fumPrice(sides.SELL))
 
             const fumBalance = (await fum.balanceOf(user1))
-            const targetFumBalance = oneEth.mul(priceWAD).div(WAD) // see "allows minting FUM" above
+            const targetFumBalance = wadMul(oneEth, priceWAD) // see "allows minting FUM" above
```
albertocuestacanada, line 195: Parameterize, please

albertocuestacanada (changes requested):
Great work. Minor changes suggested in the smart contract. I would like to request to parameterize the tests, since we are more than half way there.

Making the tests depend only on initial variables such as the initial USM mint and the oracle price will allow us to run automated tests for any scenario.

contracts/IUSM.sol
```diff
@@ -14,7 +16,15 @@ interface IUSM {
     function ethPool() external view returns (uint);
```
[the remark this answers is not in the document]
jacob-eliosoff: Yup OK done.

contracts/USM.sol
```diff
@@ -67,11 +80,17 @@ contract USM is IUSM, ERC20Permit, Delegable {
...
     {
-        uint ethOut = usmToEth(usmToBurn);
+        // First calculate:
+        uint ethOut;
+        uint ethPoolShrinkFactor;
+        (ethOut, ethPoolShrinkFactor) = ethFromBurn(usmToBurn);
+
+        // Then update state:
         _burn(from, usmToBurn);
+        _updateMintBurnAdjustment(mintBurnAdjustment().wadDiv(ethPoolShrinkFactor.wadMul(ethPoolShrinkFactor)));
```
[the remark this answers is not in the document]
jacob-eliosoff: Done

contracts/USM.sol
```diff
@@ -148,21 +179,149 @@ contract USM is IUSM, ERC20Permit, Delegable {
...
+
+    /**
+     * @notice The current min FUM buy price, equal to the stored value decayed by time since minFumBuyPriceTimestamp.
+     */
+    function minFumBuyPrice() public override view returns (uint) {
+        if (minFumBuyPriceStored == 0) {
+            return 0;
         }
-        return theoreticalFumPrice;
+        uint numHalvings = WAD * (block.timestamp - minFumBuyPriceTimestamp) / MIN_FUM_BUY_PRICE_HALF_LIFE;
```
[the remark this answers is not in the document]
jacob-eliosoff: Not using `SafeMath` here was actually just an omission!  (I noticed this earlier tonight before reading your comment here.)  Fixed.

contracts/USM.sol
```diff
@@ -148,21 +179,149 @@ contract USM is IUSM, ERC20Permit, Delegable {
...
+     * @notice The current mint-burn adjustment, equal to the stored value decayed by time since mintBurnAdjustmentTimestamp.
+     */
+    function mintBurnAdjustment() public override view returns (uint) {
+        uint numHalvings = WAD * (block.timestamp - mintBurnAdjustmentTimestamp) / BUY_SELL_ADJUSTMENTS_HALF_LIFE;
+        uint decayFactor = numHalvings.wadHalfExp(10);
+        // Here we use the idea that for any b and 0 <= p <= 1, we can crudely approximate b**p by 1 + (b-1)p = 1 + bp - p.
+        // Eg: 0.6**0.5 pulls 0.6 "about halfway" to 1 (0.8); 0.6**0.25 pulls 0.6 "about 3/4 of the way" to 1 (0.9).
+        // So b**p =~ b + (1-p)(1-b) = b + 1 - b - p + bp = 1 + bp - p.
+        // (Don't calculate it as 1 + (b-1)p because we're using uints, b-1 can be negative!)
+        return WAD + mintBurnAdjustmentStored.wadMul(decayFactor) - decayFactor;
```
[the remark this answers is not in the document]
jacob-eliosoff: Not using `SafeMath` here was again just an omission - I now understand better why/when to use it, thanks.  Patched to use safe ops consistently.

contracts/USM.sol
```diff
@@ -148,21 +179,149 @@ contract USM is IUSM, ERC20Permit, Delegable {
...
+     * @param ethIn The amount of ETH passed to mint().
+     * @return The amount of USM, and the factor by which the ethPool grew (eg, 1.2 * WAD = 1.2x).
+     */
+    function usmFromMint(uint ethIn) public override view returns (uint, uint) {
+        // Mint USM at a sliding-up USM price (ie, at a sliding-down ETH price).  **BASIC RULE:** anytime pool_eth changes by
+        // factor k, eth_price changes by factor 1/k**2.  (Earlier versions of this logic scaled price by 1/k, not 1/k**2; but
+        // that results in calls to log()/exp().)
+        uint initialUsmPrice = usmPrice(Side.Buy);
+        uint _ethPool = ethPool();
+        uint ethPoolGrowthFactor = (_ethPool + ethIn).wadDiv(_ethPool);
```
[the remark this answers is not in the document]
jacob-eliosoff: Yeah but probably best practice to just use `SafeMath` ops consistently - this was again just an oversight/misunderstanding, patched now.

contracts/USM.sol
jacob-eliosoff: Ditto.  There should now be no remaining naked `+` or `-` ops.

contracts/USM.sol
jacob-eliosoff: I patched it anyway though because why not?  Seems best to just use `SafeMath` everywhere rather than try to analyze each case.  No?

test/03_USM.test.js
jacob-eliosoff: 👍  Good call, done

test/03_USM.test.js
jacob-eliosoff: Done.

test/03_USM.test.js
jacob-eliosoff: Done

jacob-eliosoff (commented):
The change to structs looks good.  In fact now that you've done this I'm not sure my original plan (for later) of packing all the adjustments + timestamps into one 32-byte word is worth it, because most ops only update one of these pairs at a time anyway!  So updating them all at once wouldn't save much gas.  The exception (at least in the current code) is `fund()`, which updates `fundDefundAdjustmentStored` and, sometimes, `minFumBuyPriceStored`: packing everything into one word would cut those two `SSTORE`s `fund()` does to one, and wouldn't really hurt much (not much code), so maybe it is worth it - but your struct change has already delivered most of the savings.

test/03_USM.test.js
jacob-eliosoff: I don't think these parameterization requests should be too hard but enough for today.  Aiming to pick 'em up in the next couple days

jacob-eliosoff: I made some changes to test 3 to further parameterize it, a little more painful than I expected.  We could go further but maybe that's enough until we're actually putting it through random inputs.

test/03_USM.test.js
jacob-eliosoff: Done.

test/03_USM.test.js
jacob-eliosoff: Done.  At least I think so...

test/03_USM.test.js
jacob-eliosoff: Yeah done

test/03_USM.test.js
jacob-eliosoff: Hmmm, I guess this part I haven't yet parameterized.  What I've done so far is this:
1. user1 and user2 both start with certain ("parameterized") amounts of ETH.
2. user1 spends all their ETH `fund()`ing (with the FUM going to user2).
3. user2 spends all their ETH `mint()`ing (with the USM going to user1).

But this doesn't parameterize everything: eg, the `3/2` here follows because this particular test calls `fund()` twice (since the first `fund()` is a special case that doesn't use sliding prices, which we do want to test!), and I originally just made the test call both times with the same input amount of ETH.  If the second `fund()` call passed in a different amount of ETH from the first, this `3/2` would change.  So I guess it would be a matter of adding a third parameter - not just "how much ETH user1 passes to `fund()`" and "how much ETH user2 passes to `mint()`", but also "how much ETH user1 passes to their _second_ call to `fund()`"?

If that makes sense to you, I'll add that third parameter.  The math actually isn't bad.

PR #36, opened 2020-09-29
feat: usm to fum and reverse functions drafted - Fixes #23

### LINKED ISSUE #23: Fund/Defund using USM
Implement in `Proxy` by calling:
```
function fundWithUSM(...) {
    proxy.burn(...);
    proxy.fund(...);
}

function defundToUSM(...) {
    proxy.defund(...);
    proxy.mint(...);
}
```

Note that we are calling the `Proxy` functions, meaning that limit orders are included.

albertocuestacanada:
Simple conversion functions, not sure how useful they are. I'll code the tests if we want to merge this PR.

### DIFF
--- a/contracts/Proxy.sol
+++ b/contracts/Proxy.sol
@@ -27,7 +27,7 @@ contract Proxy {
     /// @dev The WETH9 contract will send ether to Proxy on `weth.withdraw` using this function.
     receive() external payable { }
 
-    /// @dev Users use `mint` in Proxy to post ETH to USM (amount = msg.value), which will be converted to Weth here.
+    /// @dev Users use `mintWithEth` in Proxy to post ETH to USM (amount = msg.value), which will be converted to Weth here.
     /// @param to Receiver of the minted USM
     /// @param minUsmOut Minimum accepted USM for a successful mint.
     function mintWithEth(address to, uint minUsmOut)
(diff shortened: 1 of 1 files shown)


alexroan:
Hmmm, I agree that they're simple, but I'm not certain they fall within our concern.

On the other hand, they're in the Proxy, not USM, so it's still a degree of separation between them. 

My worry is that if we start introducing functions like these we're blurring the line between what we set out to create, and trying to imitate platforms that exist for this exact reason (i.e, uniswap to swap between two assets). Do we want to enable a few functions that could distract from what we are trying to achieve in the first place? and prompt us to maybe blur that line even more in future?

Open question 🤷‍♂️

albertocuestacanada:
I'm not too excited about this either, @jacob-eliosoff?

jacob-eliosoff:
See my comments on the [other thread](https://github.com/alexroan/USM/issues/37#issuecomment-701216192).  I think we can probably live without `buy`-type functions, but if we do find cause to add them I don't foresee any problems.

Note that buying is a bit fundamentally distinct from minting/creating (and similarly selling vs burning/redeeming): the latter changes the total qty outstanding, the former doesn't.  Eg, if USM gets anywhere I expect markets like ETH/USM to appear on Uniswap & other exchanges: that would be real buying & selling.  Quite likely it'll be cheaper in many cases than minting/burning via our smart contract, but it's our contract's operations that keep the peg permissionless.

So in short, I wouldn't push this until we hit a real-world need.

albertocuestacanada:
Agreed. I would even go further with adding features to `Proxy`. I would just add the features that we need to have a certain behavior on the UI.

Anyone can build their own proxy, and if we are successful people will do for their own specific use cases, we don't need to do it for them.

PR #38, opened 2020-09-30
feat: No transfers to deployed contracts

albertocuestacanada:
No more oops!

### DIFF
--- a/contracts/FUM.sol
+++ b/contracts/FUM.sol
@@ -12,7 +12,19 @@ import "erc20permit/contracts/ERC20Permit.sol";
  */
 contract FUM is ERC20Permit, Ownable {
 
-    constructor() public ERC20Permit("Minimal Funding", "FUM") {
+    address public usm;
+
+    constructor(address usm_) public ERC20Permit("Minimal Funding", "FUM") {
+        usm = usm_;
+    }
+
+    /**
+     * @notice Regular transfer, disallowing transfers to this contract.
+     */
+    function transfer(address recipient, uint256 amount) public virtual override returns (bool) {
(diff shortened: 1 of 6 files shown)


jacob-eliosoff (commented):
I don't know how you do this stuff so quickly: it took me 30 min just to switch git to the right fucking branch...  I see you're checking the USM and FUM contracts, and that they reject ETH, USM and FUM sends (transfers): awesome.  My one question is, the only combo I don't see you testing is sending ETH to the FUM contract.  Is it worth adding a test like the second one here, or is there a nicer way to test it, or is testing this not worth it?

```
      it("refuses ether transfers", async () => {
        await expectRevert(
          web3.eth.sendTransaction({ from: user1, to: usm.address, value: 1 }),
          "Don't send ether here"
        )
        await expectRevert(
          web3.eth.sendTransaction({ from: user1, to: fum.address, value: 1 }),
          "Transaction reverted: function selector was not recognized and there's no fallback nor receive function"
        )
      })
```

Anyway basically this all looks great to me, good to push, thank you.

jacob-eliosoff:
Or I guess another option might (?) be to add `receive () external payable { revert("Don't send ether here"); }` to FUM as well as USM?

jacob-eliosoff:
Ah yeah that works too.  A bit more code (the extra `receive()` fn), a bit cleaner error message in the test - either way, I can push either one to this branch, up to you.

albertocuestacanada:
I realized that starting with solidity 0.6 you can't send ether to a contract unless a `receive()` function is included, so the one I added was redundant, and you can't send ether to FUM, either.

PR #46, opened 2020-10-19
CompositeOracle, MakerOracle, ...

jacob-eliosoff:
I've added `CompositeOracle` and `MakerOracle` but not yet tested them.  Not 100% sure how to test them (external calls to existing public contracts) but I may give it a poke...  The Maker "medianizer" has been pretty widely used but it has some disadvantages: seems to update only hourly (see below), more centralized than Chainlink, and I think they're planning to replace it with a whitelist-access-only version for v2 or something.  Still, it's established enough (and independent from Chainlink/Uniswap) that it's a possible ingredient in a composite oracle, and the code (from https://studydefi.com/read-maker-medianizer/) is dead simple.

@alexroan - am I right in understanding that the mainnet ETH/USD Chainlink contract is [0x5f4eC3Df9cbd43714FE2740f5E3616155c5b8419](https://etherscan.io/address/0x5f4eC3Df9cbd43714FE2740f5E3616155c5b8419)?  I found that at https://docs.chain.link/docs/reference-contracts, but I was surprised to see the last tx on there is 73 days old.  I'm guessing this contract points at others (maintained by individual source oracles), which are the ones that actually update?  Do you know how frequently?  From the chart at https://feeds.chain.link/eth-usd looks like the answer is "hourly, or more frequently when price is bouncing around," which seems reasonable.  From the txs on the [mainnet Maker medianizer](https://etherscan.io/address/0x729d19f657bd0614b4985cf1d82531c67569197b) it looks it updates just "hourly, regardless of volatility," which is considerably worse for our purposes.

It would be great to have Chainlink + Uniswap v2 + at least one other decent oracle implemented.  Then we'd be basically ready to test, make some decisions, audit & deploy.

### DIFF
--- a/buidler.config.js
+++ b/buidler.config.js
@@ -12,7 +12,7 @@ module.exports = {
       }
     },
     solc: {
-      version: "0.6.10",
+      version: "0.6.6",
       optimizer: {
         enabled: true,
         runs: 200
--- a/contracts/Delegable.sol
+++ b/contracts/Delegable.sol
@@ -1,0 +1,96 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.6.6;
+
+
+/// @dev Delegable enables users to delegate their account management to other users.
+/// Delegable implements addDelegateBySignature, to add delegates using a signature instead of a separate transaction.
+contract Delegable {
+    event Delegate(address indexed user, address indexed delegate, bool enabled);
+
+    // keccak256("Signature(address user,address delegate,uint256 nonce,uint256 deadline)");
+    bytes32 public immutable SIGNATURE_TYPEHASH = 0x0d077601844dd17f704bafff948229d27f33b57445915754dfe3d095fda2beb7;
+    bytes32 public immutable DELEGABLE_DOMAIN;
+    mapping(address => uint) public signatureCount;
+
+    mapping(address => mapping(address => bool)) public delegated;
+
+    constructor () public {
+        uint256 chainId;
+        assembly {
+            chainId := chainid()
+        }
+
+        DELEGABLE_DOMAIN = keccak256(
+            abi.encode(
+                keccak256('EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)'),
+                keccak256(bytes('Yield')),
+                keccak256(bytes('1')),
+                chainId,
+                address(this)
+            )
+        );
+    }
+
+    /// @dev Require that msg.sender is the account holder or a delegate
+    modifier onlyHolderOrDelegate(address holder, string memory errorMessage) {
+        require(
+            msg.sender == holder || delegated[holder][msg.sender],
+            errorMessage
+        );
+        _;
+    }
+
+    /// @dev Enable a delegate to act on the behalf of caller
+    function addDelegate(address delegate) public {
+        _addDelegate(msg.sender, delegate);
+    }
+
+    /// @dev Stop a delegate from acting on the behalf of caller
+    function revokeDelegate(address delegate) public {
+        _revokeDelegate(msg.sender, delegate);
+    }
+
+    /// @dev Add a delegate through an encoded signature
+    function addDelegateBySignature(address user, address delegate, uint deadline, uint8 v, bytes32 r, bytes32 s) public {
+        require(deadline >= block.timestamp, 'Delegable: Signature expired');
+
+        bytes32 hashStruct = keccak256(
+            abi.encode(
+                SIGNATURE_TYPEHASH,
+                user,
+                delegate,
+                signatureCount[user]++,
+                deadline
+            )
+        );
+
+        bytes32 digest = keccak256(
+            abi.encodePacked(
+                '\x19\x01',
+                DELEGABLE_DOMAIN,
+                hashStruct
+            )
+        );
+        address signer = ecrecover(digest, v, r, s);
+        require(
+            signer != address(0) && signer == user,
+            'Delegable: Invalid signature'
+        );
+
+        _addDelegate(user, delegate);
+    }
+
+    /// @dev Enable a delegate to act on the behalf of an user
+    function _addDelegate(address user, address delegate) internal {
+        require(!delegated[user][delegate], "Delegable: Already delegated");
+        delegated[user][delegate] = true;
+        emit Delegate(user, delegate, true);
+    }
+
+    /// @dev Stop a delegate from acting on the behalf of an user
+    function _revokeDelegate(address user, address delegate) internal {
+        require(delegated[user][delegate], "Delegable: Already undelegated");
+        delegated[user][delegate] = false;
+        emit Delegate(user, delegate, false);
+    }
+}
--- a/contracts/FUM.sol
+++ b/contracts/FUM.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
-pragma solidity ^0.6.7;
+pragma solidity ^0.6.6;
 
 import "@openzeppelin/contracts/access/Ownable.sol";
 import "erc20permit/contracts/ERC20Permit.sol";
--- a/contracts/IUSM.sol
+++ b/contracts/IUSM.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
-pragma solidity ^0.6.7;
+pragma solidity ^0.6.6;
 
 
 interface IUSM {
--- a/contracts/Proxy.sol
+++ b/contracts/Proxy.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
-pragma solidity ^0.6.7;
+pragma solidity ^0.6.6;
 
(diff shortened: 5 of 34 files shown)


jacob-eliosoff:
Btw re Maker: as of [this Sep](https://forum.makerdao.com/t/eth-oracle-question-peek-call-returns-extremely-large-value/4249/2) "the V1 Medianizer has been deprecated" and there's an application we could submit to request our contract getting whitelisted for v2 access, which I'm not too excited about

jacob-eliosoff:
https://oracles.club/ has a pretty good list of existing (on-chain?) ETH/USD oracles, with some pros & cons.

jacob-eliosoff:
Also a decent read, though no real new gems: https://medium.com/kuucrypto/will-the-real-decentralized-oracles-please-stand-up-2b936a2349c7

alexroan:
That address is a proxy. The underlying aggregator it uses is: https://etherscan.io/address/0x00c7a37b03690fb9f41b5c5af8131735c7275446

If the price fluctuates more than 0.5%, it triggers a new round, yes. Otherwise, it's hourly.

jacob-eliosoff:
Got it, thanks.  Sensible policy.

jacob-eliosoff:
I've added some stuff here (Compound "Open Price Feed" oracle, Uniswap spot oracle, support for arbitrary number of source oracles in CompositeOracle) but it's super raw...  Can't even get it to compile right now.  I may need a little help:

image

Some good news though (I think) on Uniswap - I was discussing with Patrick McCorry & Dan Robinson and they pointed out that rather than using the fancy historical TWAP with its fancy "storage proof", we could just grab the spot price.  This is considered non-kosher since the instantaneous price can be manipulated, but as Dan said, "as part of a median, and on a sufficiently liquid pair like ETH/USDC or ETH/DAI, I could see it."  I agree - I think for our purposes (part of a median) spot price is arguably _better_ than historical price anyway (more up to date).  Anyway I haven't tested it but it should be pretty easy to get working.

Perhaps we could even do something like a median-of-7:
- Chainlink ETH/USD (2x/3x weighted)
- Compound ETH/USD Open Price Feed (2x/1x weighted)
- Uniswap ETH/DAI spot
- Uniswap ETH/USDT spot
- Uniswap ETH/USDC spot

Depending on gas cost of course.  Alternatively median-of-5:
- Chainlink ETH/USD (2x weighted)
- Compound ETH/USD Open Price Feed
- Uniswap ETH/DAI spot
- Uniswap ETH/USDT spot

Or median-of-3:
- Chainlink ETH/USD
- Compound ETH/USD Open Price Feed
- Uniswap ETH/USDT (or ETH/DAI?) spot

There's also the possibility of applying for permissioned access to the MakerDao v2 medianizer, though I think the three above may be sufficient.

albertocuestacanada:
1. Make this compile.
2. Make it work with mock oracles.
3. Fork mainnet to local, and test it works there (oracles won't update their data)
4. Test in prod

albertocuestacanada:
It was quite an oversight of uniswap to leave their SafeMath.sol locked to the 0.6.6 compiler.

I downgraded all of our code to 0.6.6, and renamed a contract or two, now it compiles.

Once we get it all in order, we can upgrade the compiler again if we want to. Right now it all seems a bit messy.

albertocuestacanada:
Organized oracle contracts a bit, and did the test for the Maker Oracle. See if you can follow the pattern to test the other oracles, before we get to testing `CompositeOracle`.

Basically, each one of our oracle wrappers should take an address to the real third party oracle. When testing each oracle, you just need to do a mock contract with the same function as the real deployed oracle, but that just returns a value that we previously set.

Once all that is working, forking mainnet will show if we coded the mock contracts right.

jacob-eliosoff:
Just a quick note - the Maker v1 medianizer referenced here is now deprecated.  So don't spend too much time on that...  The v2 medianizer is a possible option, but it's whitelist-controlled (we'd need to apply to be added) - doesn't necessarily rule it out but not the first thing we should try testing I guess.

albertocuestacanada:
It doesn't matter, took me 10 minutes, it's just to show the pattern on how to test the oracles.

jacob-eliosoff:
I'm off down a rabbit hole with a possible improvement/simplification to the buy/sell adjustments, but I'll get back to oracle testing in the next ~3 days...

albertocuestacanada:
Lol, that's OK. I might do the testing myself then. Enjoy your adventure 🐰

albertocuestacanada:
I could test the Maker, Compound and Chainlink oracles, but the UniswapV2 one is difficult to create mocks for, due to the way it finds the pair (using CREATE2, instead of taking the address as a parameter).

While the UniswapV2 oracle gets tested (maybe only in testnets), I've also tested the CompositeOracle. It would be interesting to put there more than 3 oracles and do somefuzzing to make sure the sort and median work, but probably is not that necessary.

I also added one function (through a mock) to call `latestPrice` in the CompositeOracle embedded in a transaction. Not sure if it's working alright, because I was expecting something like 30K gas (of which 21K are the transaction fee), and I got 61K instead.

```
·----------------------------------------------------|---------------------------|-------------|----------------------------·
|                Solc version: 0.6.6                 ·  Optimizer enabled: true  ·  Runs: 200  ·  Block limit: 9500000 gas  │
·····················································|···························|·············|·····························
|  Methods                                                                                                                  │
······························|······················|·············|·············|·············|··············|··············
|  Contract                   ·  Method              ·  Min        ·  Max        ·  Avg        ·  # calls     ·  eur (avg)  │
······························|······················|·············|·············|·············|··············|··············
|  MockChainlinkAggregatorV3  ·  set                 ·          -  ·          -  ·      41419  ·           5  ·          -  │
······························|······················|·············|·············|·············|··············|··············
|  MockCompositeOracle        ·  latestPriceWithGas  ·          -  ·          -  ·      61672  ·           1  ·          -  │
······························|······················|·············|·············|·············|··············|··············

```

albertocuestacanada:
And I still think this has enough complexity and enough of its own set of requirements that should be moved to its own repo.

jacob-eliosoff:
Thanks Alberto!

For testing purposes (or even in prod) we could probably just pass in the Uniswap token addresses explicitly, as we do for the other oracles.  I included the tokens' mainnet addresses in the constructor comments.

We can probably get rid of the `sort()` implementation and just go back to my simple earlier `median()` code special-cased to 3 inputs, at least for now.  What I currently envision is a composite of (Chainlink, Compound, Uniswap), where the Uniswap member is itself a composite of (ETH/USDT, ETH/USDC, ETH/DAI).  So anyway let's not worry about fuzzing `sort()` yet.

61k gas sounds disturbing.  I'm pretty sure both Chainlink and Uniswap will be efficient - just reading a stored data field or two.  Hopefully the 61k is an artifact of some kind but if it turns out Compound is genuinely expensive we may have to do without it...

And yeah, totally agreed on a separate repo - the composite oracle is quite self-contained, there's no reason someone else couldn't use it once we get it working.

If the composite gets too hairy either in terms of complexity, gas or just dev time we don't need to use it, I'm not totally wedded to it.  But I do want us to at least try to get it working so we can make an informed decision.  Most of the work here is just getting the individual oracles to work which I think is useful anyway.

albertocuestacanada:
The 61K gas thing is a completely unexpected result, and Probably deserves some investigation.
 - Any loops in the system wouldn't go for longer than three iterations, having just three oracles.
 - The gas use can't come from modifying state variables, since these are all `view` functions.
 - External contract calls are 1.5K gas, I think, but we should have only 6 of them (Composite to each oracle wrapper, and each oracle wrapper to real oracle).

All in all, gas should be between 35K and 40K, of which the actual oracle reading would be 14K to 19K. I'm not sure where the extra 20K gas are going.

jacob-eliosoff:
I pushed an update to `CompositeOracle` simplifying it (eg, returning to hardcoding the number of sources as 3).  That brings the "compositeOracle.latestPrice() cost" from the test down from **40,533** before to **23,619** now.  I don't think we're going to easily get it below this (the actual `median()` call is only eating ~200 gas), but let me know if any of the numbers still look concerning.

I think we mostly just need to get Uniswap working...  Hoping to get to it in the next couple days but we'll see...

jacob-eliosoff:
(The sum of the reported `latestPrice()` gas costs of our three current individual oracles is 16,355, so that's the Composite's theoretical floor, not counting gas for wrapper contracts, normalizing decimal places, etc.)

jacob-eliosoff:
OK, `OurUniswapV2SpotOracle` and `MockUniswapV2Pair` are working now (ie with mock data only).  Next (near final!) step is to try the "fork mainnet to local" thing.  And also to get this (the composite only?  Or all our oracle code?) into its own repo.

albertocuestacanada:
Great job on getting this to work. I do not think that 25K gas for the oracle call is that bad.

Let's merge this for now, extracting the oracle to another repo can be done later. I'll also do the `fork-mainnet` feature in a different PR, and if there are any inconsistencies in using the oracles we can fix them there.

albertocuestacanada (approved):
Let's merge for now. The code is solid enough and this is a huge change already.

PR #49, opened 2020-10-26
Merge mintBurnAdj & fundDefundAdj into buySellAdj

### LINKED ISSUE #48: Combine mintBurnAdjustment and fundDefundAdjustment into a single buySellAdjustment
This is an idea I've had in the back of my mind, to fix some unpleasant cases where the system charges both buyers and sellers an excessive fee.  I dove in this week and looks like it's working - it's a significant change (different formulas in `usmFromMint()` and co), but a worthwhile one I believe.  Doesn't change the public API at all.  I'm putting some time into revamping the tests before I submit a PR - aiming for sometime tomorrow.  Then I can get back to oracle work!

jacob-eliosoff:
OK, finally got the tests working, and attempted a cleanup of useful-but-hairy `03_USM.test.js`.  Please have a look and I hope to merge this soon!  Very happy to answer any questions though.  Also any suggestions on test structure/formatting welcome.  This closes #48.


test/03_USM.test.js
```diff
@@ -255,186 +212,369 @@ contract('USM', (accounts) => {
...
-            // - Use timeMachine to move forward 180 secs = 3 min.
-            // - Since USM.BUY_SELL_ADJUSTMENTS_HALF_LIFE = 60 (1 min), 3 min should mean a decay of ~1/8.
-            // - The way our "time decay towards 1" approximation works, decaying 1/4 by ~1/8 should yield 1 + (1/4 * 1/8) - 1/8 = 0.90625.
-            //   (See comment in USM.mintBurnAdjustment().)
-            // - And finally - since our decay's exponentiation is approximate (WadMath.wadHalfExp()), don't compare mintBurnAdj3 to
-            //   targetMintBurnAdj3 *exactly*, but instead to within some reasonable error (eg, 10).
-            const mintBurnAdj2 = (await usm.mintBurnAdjustment())
-            
-            const block0 = (await web3.eth.getBlockNumber())
+          it("uses flat USM price first time USM is minted", async () => {
```
albertocuestacanada, line 215: Does this test anything?

test/03_USM.test.js
```diff
@@ -255,186 +212,369 @@ contract('USM', (accounts) => {
...
+              usmSellPrice3 = await usm.usmPrice(sides.SELL)
+            })
+
+            it("slides price correctly when minting FUM", async () => {
+              const targetEthPool3 = ethPool2.add(ethPerFund)
+              shouldEqual(ethPool3, targetEthPool3)
+
+              // Check vs the integral math in USM.fumFromFund():
+              const integralFirstPart = wadMul(debtRatio2, wadSquared(ethPool3).sub(wadSquared(ethPool2)))
+              const targetFumBalance3 = wadDiv(wadSqrt(integralFirstPart.add(wadSquared(wadMul(totalFumSupply1, fumBuyPrice2)))),
```
albertocuestacanada, line 293: If you use the same `wadSqrt` implementation in both the contract and the tests, then you should be sure beyond doubt the algorithm works. Are there tests for it anywhere?

albertocuestacanada (approved):
Mostly minor comments, it's beyond my capacity to assess the content of te smart contract or the tests. In general the code quality is great.

I suggest that we work on invariants if we want to know this does what it should. Some article explaining what is going on and getting feedback from the community would help as well.

test/03_USM.test.js
jacob-eliosoff: Well, it tests that the first time `mint()` is called, the price per USM is just linear (`wadMul(ethPerMint, priceWAD)` a few lines down) rather than on a sliding scale, as it will be for all future calls.  This is a "make sure things are working as expected" test rather than a "make sure something bad doesn't happen" test.

test/03_USM.test.js
jacob-eliosoff: No, good point.  The code was lifted from production Uniswap (specifically from https://ethereum.stackexchange.com/a/87713) and I tested it pretty thoroughly, but only on an ad hoc basis - I'll add a proper test.

jacob-eliosoff: Added.  (Have a look if you like, I snuck it into `03_USM.test.js`.)

PR #50, opened 2020-10-28
Integrate USM with CompositeOracle

albertocuestacanada:
- Check that the tests that are now allowed some tolerance should do so.
- Check if there is a way of making the commented out tests in `06_USM_Oracle.js` work. I think they just require to implement `setPrice` functions in the MockOracles and to work out the right calls to bring the price where we want it to be for the test scenarios to be right.

Also set up the migration scripts and mainnet forking scripts.

To deploy to a local mainnet fork.
0. Create a `.secret` file in the project root with this text inside: `how are you gentlemen all your mnemonic are belong to us`
1. In one terminal run `yarn mainnet-ganache`
2. In another terminal run `npx truffle migrate --reset --network mainnet-ganache --skip-dry-run`

To connect to a deployed instance.
`npx truffle console --network mainnet-ganache`

Once in the console, you can connect to the deployed contracts as in:
`weth = await WETH9.deployed()`
`pair = await IUniswapV2Pair.at('0x0d4a11d5eeaac28ec3f61d100daf4d40471f1852')`

It's javascript, so you can execute the same commands as in the tests to execute stuff. Some extra useful commands:
`accounts = await web3.eth.getAccounts()`
`ethBalance = await web3.eth.getBalance(accounts[0])`


jacob-eliosoff:
Clearly the 90% redundancy between test 3 and test 6 isn't ideal.  I have no simple suggestions right now though

PR #51, opened 2020-11-02
feat: merge oracles

albertocuestacanada:
:vomiting_face: 

But `compositeOracle.latestPrice` goes down from 26K to 17K.


alexroan (approved):
I hate it, but it's a huge saving of GAS.

PR #52, opened 2020-11-02
Experiment: move all oracle functionality into USM to save gas?

jacob-eliosoff:
This is experimental and incomplete - eg right now test 3 is hacked to just use three copies of the same Uniswap pair.  I just wanted to get a sense of how much gas this could save and how ugly the code would be...  It saves a decent chunk of gas (eg `mint()` is 211k in master vs 172k here - see screenshots below), and the code isn't really super ugly, but it is much less modular & flexible: eg hardcoding Chainlink/Compound/Uniswap oracles in `USM` (well, in `Oracle`, which `USM` now _inherits_ from - so no longer pluggable).

I dunno, I guess I think this approach is reasonable assuming we don't find anything it breaks horrendously.  Fwiw getting test 3 to pass here was much easier than I expected.  I wrote very little code here, it's mostly just shifted around.

Test 3 on this branch:
image

vs master:
image


jacob-eliosoff:
I just pushed a further restructuring that may be somewhat cleaner, eg Chainlink/Compound/Uniswap each in their own file...  But maybe not in which case we should just revert it.  Also test 3 is failing now and it's a bit beyond me:

```
  1) Contract: USM
       mints and burns a static amount
         "before each" hook for "starts with correct FUM price":
     Error: MockUSM contains unresolved libraries. You must deploy and link the following libraries before you can deploy a new version of MockUSM: $e0fb37bf338e0ecbaef71b67f51829a4d9$
      at Context.<anonymous> (test\03_USM.test.js:125:26)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

Googling led me to an apparently well-known [deployment/migration subtlety](https://ethereum.stackexchange.com/questions/39372/you-must-deploy-and-link-the-following-libraries-before-you-can-deploy-a-new-ver) but I failed to overcome it...

alexroan:
I like this, gonna dig into it shortly!

alexroan:
@jacob-eliosoff looks like we have some conflicts since #51 went into master.

Can you resolve those?

alexroan (changes requested):
Looks like we have some conflicts since #51 went into master.

Can you resolve those please? :)

jacob-eliosoff:
Yeah first I'm gonna get the other tests working then merge #51 into this repo.  Sometime today

jacob-eliosoff:
I would say hold off on review till I fix those tests + merge, thank you!

jacob-eliosoff:
OK, I closed the remaining ~6k gap between my old code above's 172k gas for `mint()` and today's 178k: back down to 172k.  Also merged with master/#51, though I think I basically squashed your changes @albertocuestacanada - if you see any that should have been preserved here and weren't, let me know...

With that, this PR is now ready for review!  Aiming to merge in the next day or two.  We're quite close to audit submission now I think.  Maybe we should get it up on a testnet first and play with it a bit?

PR #53, opened 2020-11-11
Fixed integral math in usmFromMint() etc

jacob-eliosoff:
Afaict the old math was just wrong and this fixes it.  When I get a chance in the next couple days I'll do more testing via forked mainnet and we'll see if I find any more bugs - feel free to hold off on reviewing this till I report back.

### DIFF
--- a/contracts/WadMath.sol
+++ b/contracts/WadMath.sol
@@ -28,7 +28,7 @@ library WadMath {
     }
 
     function wadCubed(uint x) internal pure returns (uint) {
-        return (((x.mul(x)).add(WAD_OVER_2) / WAD).mul(x)).add(WAD_OVER_2) / WAD;
+        return ((x.mul(x)).add(WAD_OVER_2).mul(x)).add(WAD_OVER_2) / WAD_SQUARED;
     }
 
     //rounds to zero if x/y < WAD / 2
--- a/contracts/WadMath.sol
+++ b/contracts/WadMath.sol
@@ -10,32 +10,55 @@ import "@openzeppelin/contracts/math/SafeMath.sol";
 library WadMath {
     using SafeMath for uint;
 
-    uint private constant WAD = 10 ** 18;
+    enum Round {Down, Up}
 
-    uint private constant WAD_OVER_2 = WAD / 2;
+    uint private constant WAD = 10 ** 18;
+    uint private constant WAD_MINUS_1 = WAD - 1;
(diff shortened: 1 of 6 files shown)


contracts/WadMath.sol
```diff
@@ -76,19 +81,15 @@ library WadMath {
...
+    function wadCbrt(uint y) internal pure returns (uint root) {
+        if (y > 0 ) {
+            y = y.mul(WAD_SQUARED);
+            uint newRoot = y.add(2) / 3;
+            do {
+                root = newRoot;
+                newRoot = (root + root + ((y / root) / root)) / 3;
+            } while (newRoot < root);
         }
+	//require(root ** 3 <= y && (root + 1) ** 3 > y);
```
albertocuestacanada, line 93: Would this be an `assert`, or something that you would like to test by fuzzing?

albertocuestacanada (approved):
Fine by me, but we should get someone with proper math chops to review your math. Maybe we could ask in twitter for volunteers?

contracts/WadMath.sol
jacob-eliosoff: Yeah I think something to test by fuzzing.  This finding the cube root thing turned into another rabbit hole...  I think what I have now is fine though, very short simple code and very cheap in gas.

jacob-eliosoff:
> Fine by me, but we should get someone with proper math chops to review your math. Maybe we could ask in twitter for volunteers?

I don't know if I'll get helpful responses on that...  It's not so much "Is the math right" as "Are we calculating the right thing."  But yeah my next steps here are to do some Ganache testing of the latest code (today), then try to write a Medium update summarizing the latest math tomorrow/this weekend, and see if anyone has comments.

But if Ganache goes OK (and I don't find a bug like last time...), I'm inclined to merge this and send it off for audit, maybe even tomorrow?  I've worked through a bunch of pen-&-paper examples with the current math, I _think_ it's OK...

The summary of the current sliding-price math is: when an operation (`mint`/`burn`/`fund`/`defund`) changes debt ratio from `dr1` to `dr2`, the price adjustment changes by `(dr2/dr1)**2`.  So, eg, `funds`/`burns` decrease the debt ratio, which increases the adjustment, which hikes up the FUM buy price for further `funds` and reduces the USM sell price for further `burns`.

jacob-eliosoff:
OK I've laboriously checked a bunch of examples by hand and I think this is good to push!  Next step: audit...

PR #56, opened 2020-11-18
feat: Made Ether the underlying currency

albertocuestacanada:
Please review paying special attention to when `ethPool` is updated.

I haven't even touched the tests yet.


jacob-eliosoff:
I took a crack at this and I think fundamentally it's fine, but some glitches...

1. Test 4 is still failing, and I had to hack test 3 a bit, because I'm getting weird errors like `Invalid number of parameters for "mint". Got 6 expected 0!`  I think something about the `{ from: user1, value: oneEth }`-type part of the arguments is confusing JavaScript?...  Can you have a look?

2. Relatedly - to get test 3 to pass I had to comment out the `onlyHolderOrDelegate(from, "Only holder or delegate")` part of the fn signatures in `USMTemplate`.  (This hack didn't work for test 4 because test 4 actually needs to set `msg.value` so it can pass in ETH.)  Maybe it's obsolete anyway with Proxy gone?  (I also removed Proxy as part of these changes.)

3. I suspect that's just syntax nonsense we can figure out...  A more interesting fundamental question is, if `mint()`/`fund()` can pass in either ETH or WETH, which should the USM contract store its pool as?  Options:\
\
a) Store as ETH.  Any WETH passed in gets converted to ETH.  When a `burn()`/`defund()` caller request outputType WETH, some ETH is converted to WETH ("deposited") before it's sent to the user.  This is what this branch does now.
b) Store as WETH.  Mirror image of the above: when a caller requests ETH, it will be "withdrawn" from the WETH pool before it's sent.  This is what master currently does.
c) Store a mix of both, whatever's received!  This will save callers (esp of `mint()`/`fund()`) some gas, but it requires a tad more withdrawal logic: eg, a user requests ETH but the contract only has WETH; or they're supposed to get back 5 ETH and the contract currently holds 4 ETH + 3 WETH...\
\
I think any of these three will work really...

jacob-eliosoff:
OK I know I just pushed yet another big commit but it's just an iteration on the previous change.  After sleeping on it it seemed to me that the most elegant approach, and the one most in our spirit of minimalism, was one Alberto suggested: make `USM`/`USMTemplate` 100% ETH-only, moving all WETH logic into `Proxy`.  This cleans up the fn arguments and gets rid of a lot of `if`-`else` cases.  Also the `Proxy` fns are super compact now.

I might have missed some WETH `approves` or whatnot, I get confused about that stuff, but all the tests are passing.

And as a bonus, check out gas use now (below)!  Getting really lean.

I think this is pretty much good to merge but kindly have a look, mostly just at `Proxy` and `USMTemplate`.  The "Files changed" diffs vs master above are probably the clearest.
image

PR #57, opened 2020-11-20
Can we let users do mint/burn/fund/defund just by sending ETH/USM/FUM?

jacob-eliosoff:
This is just draft code to start from...  You may know cleaner ways.  The "send ETH to USM contract to do a `mint()`" part works here at least some of the time but "send USM to USM contract (or FUM contract) to do a `burn()`" didn't work for me just now.

### DIFF
--- a/contracts/FUM.sol
+++ b/contracts/FUM.sol
@@ -19,12 +19,20 @@ contract FUM is ERC20Permit, Ownable {
     }
 
     /**
+     * @notice If anyone sends ETH here, assume they intend it as a `fund`.
+     */
+    receive() external payable {
+        usm.fundTo{ value: msg.value }(msg.sender, 0);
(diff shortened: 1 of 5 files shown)


albertocuestacanada:
All done, check it out and merge. Feel free to play with it in a mainnet fork.

jacob-eliosoff:
> Adding `defundFromFUM()` isn't a huge deal but it is a bit of a drag... I don't see a way around it though? Unless there's a clean way to set the `FUM` as a delegate of the `USMTemplate`, but I couldn't get that to work since `usm.fum` is immutable. I could imagine scrapping its immutability to get rid of this function...

I hate `defundFromFUM()` but hey it works.  So if we don't find something cleaner I guess it stays

PR #58, opened 2020-11-22
Parsing minOut from sends

jacob-eliosoff:
I think this is good to merge but feel free to have a quick look before I do


jacob-eliosoff:
Merging but do let me know any suggestions.  Aside from the fact that this should really have been three separate commits (the `MinOut` stuff, simplifying `mint()` + `mintTo()` -> `mint()` etc, support for sends to `address(0)`) - I will work on doing better at that...

PR #59, opened 2020-11-23
feat: refactor and comments

albertocuestacanada:
Moved the parameter processing to the function that takes it, and wrote comments. Please check.


jacob-eliosoff:
Yeah I deliberated about where to put the call to `parseMinEthOut()`.  And I do normally write comments (esp for confusing stuff like this), I intentionally left this one a bit obfuscated.  But basically these changes look fine!

PR #60, opened 2020-11-24
Override _transfer() rather than transfer()?

jacob-eliosoff:
Reading the [comments](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol#L197) for OpenZeppelin's `_transfer()` fn ("This internal function is equivalent to `transfer`, and can be used to eg implement automatic token fees, slashing mechanisms, etc"), and the fact that they chose to make it `internal virtual` rather than say `private`, I think this is what they intended overriders like us to do.  Give it a think...  Tests pass fine.

### DIFF
--- a/contracts/FUM.sol
+++ b/contracts/FUM.sol
@@ -30,7 +30,7 @@ contract FUM is ERC20Permit, Ownable {
 
     /**
      * @notice If a user sends FUM tokens directly to this contract (or to the USM contract), assume they intend it as a `defund`.
-     * If using `transfer`/`transferFrom` as `defund, and if decimals 8 to 11 (included) of the amount transferred received
+     * If using `transfer`/`transferFrom` as `defund`, and if decimals 8 to 11 (included) of the amount transferred received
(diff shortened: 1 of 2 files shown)


contracts/FUM.sol
```diff
@@ -30,17 +30,15 @@ contract FUM is ERC20Permit, Ownable {
 
     /**
      * @notice If a user sends FUM tokens directly to this contract (or to the USM contract), assume they intend it as a `defund`.
-     * If using `transfer` as `defund, and if decimals 8 to 11 (included) of the amount transferred received
+     * If using `transfer`/`transferFrom` as `defund, and if decimals 8 to 11 (included) of the amount transferred received
```
albertocuestacanada, line 33:
```suggestion
     * If using `transfer`/`transferFrom` as `defund`, and if decimals 8 to 11 (included) of the amount transferred received
```

albertocuestacanada (approved):
Fair enough, it's cleaner than overriding both `transfer` and `transferFrom`.

PR #63, opened 2020-12-01
TWAP, both for one Uniswap pair and for median-of-three-pairs

jacob-eliosoff:
This definitely needs some eyeballs on it...  The key files are:
- `OurUniswapV2TWAPOracle`, which implements TWAP for a single Uniswap pair.  `MIN_TWAP_PERIOD` is currently set (a bit arbitrarily) to 2 minutes - see the long comment at the top.  `USM` doesn't actually use this file, but it's an easier-to-read-and-debug little brother of...
- `UniswapMedianTWAPOracle`, which mirrors the TWAP-storing-and-calculating logic in `OurUniswapV2TWAPOracle` but for _three_ Uniswap pairs, then taking their median (as we currently do in `UniswapMedianSpotOracle`).  In particular, the logic for storing three pairs' cumulative prices (plus timestamps) in one 256-bit word could use review...
- `MedianOracle` used to take `UniswapMedianSpotOracle` as one of its three sources to median (along with Chainlink and Compound): now I swapped in `UniswapMedianTWAPOracle` instead.  `USM` inherits `latestPrice()` from `MedianOracle`, so changing how `MedianOracle` works implicitly changes `USM` too.
- `OurUniswap` has the actual TWAP logic - that math itself isn't too messy.
- And I updated tests 01 and 03 to do some minimal checks of the TWAP logic.  In particular, to make it call `cacheLatestPrice()`, the function that records a TWAP observation so that later `latestPrice()` calls have a stored reference to price against.

The tests do pass (and I did some basic sanity-checking via logging), and I've tried to keep the code reasonably clean.  Currently `mint` is at 163k gas vs 100k in master but I expect we can get that down, I haven't tried at all yet.

### DIFF
--- a/contracts/USM.sol
+++ b/contracts/USM.sol
@@ -10,14 +10,11 @@ contract USM is USMTemplate, MedianOracle {
     constructor(
         AggregatorV3Interface chainlinkAggregator,
         UniswapAnchoredView compoundView,
-        IUniswapV2Pair[NUM_UNISWAP_PAIRS] memory uniswapPairs,
-        uint[NUM_UNISWAP_PAIRS] memory uniswapTokens0Decimals,
-        uint[NUM_UNISWAP_PAIRS] memory uniswapTokens1Decimals,
-        bool[NUM_UNISWAP_PAIRS] memory uniswapTokensInReverseOrder
+        IUniswapV2Pair uniswapPair, uint uniswapToken0Decimals, uint uniswapToken1Decimals, bool uniswapTokensInReverseOrder
     ) public
         USMTemplate()
         MedianOracle(chainlinkAggregator, compoundView,
-                     uniswapPairs, uniswapTokens0Decimals, uniswapTokens1Decimals, uniswapTokensInReverseOrder) {}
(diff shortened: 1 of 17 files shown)


jacob-eliosoff:
Fixed a couple of (related) bugs involving `cacheLatestPrice()` triggering wasteful extra calls to `latestPrice()` (and missing others).  Gas is now down to eg 124k for `mint()` - still above master's 100k, but _much_ better than the 163k I was originally seeing.

jacob-eliosoff:
These last two commits switch `USM` from using `UniswapMedianTWAPOracle` (median of three Uniswap TWAPs) to `OurUniswapV2TWAPOracle` (a single Uniswap TWAP - specifically, ETH/USDC).  As @albertocuestacanada has argued, this is simpler (less bug risk), cheaper, and seems like it offers adequate protection against source oracle problems.  It also cuts gas significantly: `mint()` down to 98k, cheaper than master.

jacob-eliosoff:
(Though that "cheaper than master" probably depends on whether your `mint()` call happens to be one of the ones that does an `SSTORE` to record a cumulative-price observation.)

jacob-eliosoff:
I've poked/tested extensively but I can't promise this is 100% bug-free, but I think it's good to merge...  Onward!

PR #72, opened 2020-12-13
feat: implemented caps

albertocuestacanada:
Max of 1000 FUM and USM per address. Max of 100 ETH pooled between all users.

Someone can sybil attack and use 50 addresses or so to use up all the liquidity, but it would be a bit pointless.


jacob-eliosoff:
I pushed a couple of proposed tweaks - optional though, feel free to revert:

1. Make the ETH cap grow gradually (2 ETH per day), starting from 40-or-so up to a final Jan 15 cap of 100 ETH.  This ensures that a rush of initial operations won't make it unusable for the rest of the month.
2. Make the cap 0 after Jan 15, so the contract becomes withdrawal-only.

PR #79, opened 2021-01-14
Feat/moving mid

jacob-eliosoff:
This is a bunch of changes related to the two big new features.  Not ready to merge or even test yet, I'll update this PR when it is.  I am trying to make each commit reasonably legible and bite-sized.

1. "Moving mid", ie, `mints`/`burns`/`funds`/`defunds` move the USM contract's mid ETH/USD price _separately_ from oracle updates.  (When a fresh oracle price comes in, the USM mid is reset to the oracle's mid; this change is about what happens to the mid _between_ oracle updates.)
\
In plain English, the behavior we're targeting here is:  "A short-ETH op (`mint`/`defund`) pushes the system's mid ETH price down, and pushes the price you actually sell at down even more."  And of course the reverse for the long-ETH ops, `burn` and `fund`.

2. New sliding-price math, based on % change in ETH pool, rather than on % change in debt ratio.  Eg, suppose the pool contains 200 ETH, and the user calls `mint(10 ETH)`, increasing the pool by 5% = a factor of 1.05×.  Then our new math says:
- The ETH mid decreases by 1/sqrt(1.05) = 0.9759×.  ("Decreases" because minting amounts to selling ETH for USD, so it pushes the ETH price down.)
- The buy-sell adjustment also decreases, by the same factor, 1/sqrt(1.05) = 0.9759×.  That is, as the mid price drops, your sell price gets further below mid, too.
- So, combined, the user's ETH sell price decreases (ie, USM buy price increases) by the product of these factors: 1/sqrt(1.05) × 1/sqrt(1.05) = 1/1.05 = 0.9524×.
\
This math is actually all straightforward: the hairy part is how to translate this for `fund`/`defund`...  More on that soon.


jacob-eliosoff:
OK, this is now feature-complete, the tests pass, and I think it's good to merge!  I need to add/update quite a few comments, and I expect there are some little bugs left, this is hot off the presses.  But I'm not aware of any gaping holes.

I propose we get this up on kovan and fool around with it (but not announce the testnet version publicly yet because it's too raw - may need to redeploy).

Also, don't deploy until reading my next comment here...

jacob-eliosoff:
See also #85, which is just this codebase but with the additional undoing of the inheritance hack.  I think #85 is what we should be merging and deploying to testnet.

PR #80, opened 2021-01-14
Test cleanup

juanpcapurro:
I was trying to understand how the project works via its test suite and found a few issues, this is roughly the list of things I changed:

- removed fuzzing tests, they had outdated contract calls and initializations and weren't being run by the `test` script
- `lint:js` now lints & fix all the test files
- removed some redundant checks for oracle prices
- modified `lint:js` so it calls `eslint` enforcing prettier rules. This has the benefit of also having a linter for use in the editor.
- running the linter revealed a lot of unused / undefined variables, fixed those. Also a few minor styiling things changed, such as which kinds of quotes are used. I could revert those if you want.

The linter configuration and automatic fixing was done in the first commit, so you can review it separately from the manual changes


deploy/1_deploy_usmfum.js
```diff
@@ -1,41 +1,33 @@
...
+  1: '0x922018674c12a7f0d394ebeef9b58f186cde13c1',
+  42: '0x0000000000000000000000000000000000000000',
 }
 const usdcEthAddresses = {
-  '1' : '0xb4e16d0168e52d35cacd2c6185b44281ec28c9dc',
-  '42' : '0x0000000000000000000000000000000000000000',
-}
-const daiEthAddresses = {
-  '1' : '0xa478c2975ab1ea89e8196811f51a7b7ade33eb11',
-  '42' : '0x0000000000000000000000000000000000000000',
```
jacob-eliosoff, line 20: I think it's worth keeping the USDT & DAI addresses around - we could need them sometime (possibly even in a hurry!).

test/01_TestOracle.test.js
```diff
@@ -18,19 +18,18 @@ const GasMeasuredOracleWrapper = artifacts.require('GasMeasuredOracleWrapper')
...
-  const chainlinkPrice = '38598000000'                      // Chainlink aggregator (AggregatorV3Interface) stores 8 dec places
+  const chainlinkPrice = '38598000000' // Chainlink aggregator (AggregatorV3Interface) stores 8 dec places
   const chainlinkPriceWAD = new BN(chainlinkPrice + '0000000000') // We want 18 dec places, so add 10 0s
 
-  const compoundPrice = '414174999'                         // Compound view (UniswapAnchoredView) stores 6 dec places
+  const compoundPrice = '414174999' // Compound view (UniswapAnchoredView) stores 6 dec places
   const compoundPriceWAD = new BN(compoundPrice + '000000000000') // We want 18 dec places, so add 12 0s
 
-  const ethDecimals = new BN(18)                            // See OurUniswapV2TWAPOracle
+  const ethDecimals = new BN(18) // See OurUniswapV2TWAPOracle
```
jacob-eliosoff, line 29: Personally I would probably leave the spacing choices on these comments (and others) as they are...  I find them easier to pick out visually.  Do you think this is bad?

test/01_TestOracle.test.js
```diff
@@ -48,50 +47,30 @@ contract('Oracle pricing', (accounts) => {
...
-  const usdcEthScaleFactor = (new BN(10)).pow(ethDecimals.add(new BN(18)).sub(usdcDecimals))
-
-  const daiEthCumPrice0_0 = new BN('291033362911607134656453476145906896216') // From the DAI/ETH pair
-  const daiEthCumPrice1_0 = new BN('30339833685805974401597062880404438583745289')
-  const daiEthTimestamp_0 = new BN('1606780728')
-  const daiEthCumPrice0_1 = new BN('291036072023637413832938851532265880018')
-  const daiEthCumPrice1_1 = new BN('30340852730044753766501469633003499944051151')
-  const daiEthTimestamp_1 = new BN('1606781048')
-  const daiEthTokensInReverseOrder = true
-  const daiEthScaleFactor = (new BN(10)).pow(ethDecimals.add(new BN(18)).sub(daiDecimals))
```
jacob-eliosoff, line 60: Again, I'd probably keep these in case we need to swap in one of the other pairs sometime, but I could be convinced!  I wonder what @albertocuestacanada thinks.

test/03_USM.test.js
```diff
@@ -14,8 +14,21 @@ require('chai').use(require('chai-as-promised')).should()
...
+    4,
+    6,
+    8,
+    18,
+    100,
+    1000,
+    '1000000000000000000',
+  ].map(function (n) {
+    return new BN(n)
+  })
```
jacob-eliosoff, line 31: I dunno this doesn't read like an improvement to me...  Vertical space is valuable!

test/03_USM.test.js
```diff
@@ -93,18 +115,14 @@ contract('USM', (accounts) => {
...
     // Check that abs(x - y) < 0.0000001(x + y):
-    const diff = (x.gt(y) ? x.sub(y) : y.sub(x))
+    const diff = x.gt(y) ? x.sub(y) : y.sub(x)
     diff.should.be.bignumber.lt(x.add(y).div(new BN(1000000)))
   }
 
-  function fl(w) { // Converts a WAD fixed-point (eg, 2.7e18) to an unscaled float (2.7).  Of course may lose a bit of precision
-    return parseFloat(w) / 10**18
-  }
-
```
jacob-eliosoff, line 103: We should hang onto this, I often use it in my `console.log()`s when debugging.

jacob-eliosoff (changes requested):
Thanks for this @juanpcapurro!

Yeah there was definitely a lot of gunk in these tests.  I'm pretty new to all this and didn't even know how to lint them...

All the stuff about removing unused vars etc is very helpful, thank you.  My objections here mostly amount to three fairly superficial/subjective things:
1. I disagree with stuff like splitting out a `let` statement with 10 vars into 10 lines, one line per var.  Imagine if we put every argument to every function on its own line...  It would greatly reduce the information density of a page of code.
2. I more tentatively disagree with removing the extra spaces before inline comments.  I do find them easier to read when they're lined up visually...  I dunno I don't feel too strongly about it
3. I favor keeping around a few things that are currently unused but could quite likely be used in the future, like the USDT & DAI addresses.

Let me know your thoughts and I'd like to see this merged!

deploy/1_deploy_usmfum.js
juanpcapurro: If that's the case we could then move them to their own `adddresses/contracts/constants` files, to both keep the linter happy and so they're reusable

test/01_TestOracle.test.js
juanpcapurro: I don't think it's bad, but it's the linter that enforces it. I could see what is the rule that forbids that and disable it if you want

test/03_USM.test.js
juanpcapurro: yeah that makes sense, we could move it into its own `debugUtils` or something file. Aside from the reasons quoted above, when first getting into the tests it was hard to figure out what is used for testing and what isn't used or has some other purpose like this one

test/03_USM.test.js
juanpcapurro: I could disable the linter rule for it or try setting a longer target line length for prettier: https://prettier.io/docs/en/options.html#print-width (if that's what formatting it this way)

test/01_TestOracle.test.js
juanpcapurro: If we need this in the future we could go back in git history or prehaps move it into a constants file

PR #89, opened 2021-01-20
WIP: Zero-fee Optional Upgrade

### LINKED ISSUE #25: 0-fee upgrade op
0-fee upgrade op, to save users money if/when we need to push a fresh contract

alexroan:
Fixes #25 .

In the event of the protocol acting unexpectedly, or to enable users to upgrade to a new version, proposal to add a zero-fee optional upgrade mechanism.

In the event of this happening, USM1 is the compromised/old contract, USM2 is the newly deployed upgrade contract.

Owner flow:
1. Deploy USM2 (with a `recover` function intended to start the USM1 holder user flow)
2. Propose USM2 address to USM1 by calling `proposeUpgradeAddress`
3. Confirm proposed address by calling `confirmUpgradeAddress`

User flow:
1. USM1 holder calls a `recover` (or some similar) function on USM2
2. USM2 calculates the amount of ETH collateral that USM1 should transfer to it along with the user's full USM1 balance, then calls `requestSwap(holder, ethAmount)` on USM1
3. `requestSwap` burns the tokens, transfers the ETH, and returns true if all is well
4. USM2 is responsible for minting USM2 tokens to the holder depending on the result of `requestSwap`

### DIFF
--- a/contracts/UpgradableUSM.sol
+++ b/contracts/UpgradableUSM.sol
@@ -50,13 +50,18 @@ contract UpgradableUSM is USM, Ownable {
      * @param ethAmount uint amount of ETH collateral to send with swap
      * @return success - This determines whether or not the upgrade contract mints new tokens to the holder
      */
-    function requestSwap(address holder, uint ethAmount) external onlyUpgradeAddress returns (bool){
+    function requestSwap(address holder, uint ethAmount) external upgrading onlyUpgradeAddress returns (bool){
         require(address(this).balance >= ethAmount, "Not enough ether");
         uint tokenBalance = balanceOf(holder);
         require(tokenBalance > 0, "Holder has no tokens to swap");
         _burn(holder, tokenBalance);
         payable(upgradeAddress).sendValue(ethAmount);
         return true;
+    }
+
+    modifier upgrading() {
+        require(upgradeAddress != address(0), "Not currently upgrading");
+        _;
     }
 
     modifier upgradeAddressNotSet() {
--- a/contracts/UpgradableUSM.sol
+++ b/contracts/UpgradableUSM.sol
@@ -1,0 +1,71 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.6.6;
+
+import "./USM.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/utils/Address.sol";
+import "./oracles/Oracle.sol";
+
+contract UpgradableUSM is USM, Ownable {
+    using Address for address;
+    using Address for address payable;
+
+    address public upgradeAddress;
+    address public proposedUpgradeAddress;
+
+    event UpgradeAddressProposed(address indexed previousProposal, address indexed newProposal);
+    event UpgradeAddressConfirmed(address indexed upgradeAddress);
+
+    constructor(Oracle oracle_) public USM(oracle_) {}
+
+    /**
+     * @notice Propose an upgrade address.
+     * @dev Must be a contract address, callable only by owner and callable only if the
+     * upgradeAddress is not already set 
+     * @param proposed address
+     */
+    function proposeUpgradeAddress(address proposed) external onlyOwner upgradeAddressNotSet {
+        require(proposed.isContract(), "Must be contract");
+        address priorProposal = proposedUpgradeAddress;
+        proposedUpgradeAddress = proposed;
+        emit UpgradeAddressProposed(priorProposal, proposed);
+    }
+
+    /**
+     * @notice Confirm the currently proposedUpgradeAddress as the upgradeAddress/
+     * @dev THIS CAN ONLY BE DONE ONCE. Only callable by owner and upgradeAddress must not already
+     * be set.
+     * @param confirmed address to confirm
+     */
+    function confirmUpgradeAddress(address confirmed) external onlyOwner upgradeAddressNotSet {
+        require(proposedUpgradeAddress == confirmed, "Does not match proposed");
+        upgradeAddress = proposedUpgradeAddress;
+        emit UpgradeAddressConfirmed(upgradeAddress);
+    }
+
+    /**
+     * @notice Request to swap all old tokens owned by the holder to new upgrade tokens
+     * @dev This must be called by the upgrade address
+     * @param holder address
+     * @param ethAmount uint amount of ETH collateral to send with swap
+     * @return success - This determines whether or not the upgrade contract mints new tokens to the holder
+     */
+    function requestSwap(address holder, uint ethAmount) external onlyUpgradeAddress returns (bool){
+        require(address(this).balance >= ethAmount, "Not enough ether");
+        uint tokenBalance = balanceOf(holder);
+        require(tokenBalance > 0, "Holder has no tokens to swap");
+        _burn(holder, tokenBalance);
+        payable(upgradeAddress).sendValue(ethAmount);
+        return true;
+    }
+
+    modifier upgradeAddressNotSet() {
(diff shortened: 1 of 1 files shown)


jacob-eliosoff:
So this raises some interesting questions.

1. First of all, on the tactical proposal: I think a "minimally intrusive" approach here would be to create the `USM` instance with two oracles, `oracle1` and `oracle2` (or even an arbitrary-length array), in addition to `usm.oracle` (ie, the oracle it actually uses), and then have an admin-controlled `setOracle(uint oracleIndex)` function which would let us switch between the two oracles in a pinch.  Assuming both preset oracles are sane, this seems like a power reasonably impossible for us to abuse.\
\
And yes, I think a plain `ChainlinkOracle` instance is a fine candidate for `oracle2`!\
\
(Note that this only works because of our recent change to make `USM` no longer _inherit_ the `Oracle` it uses!)

2. For more serious problems, where just falling back to a preloaded `oracle2` isn't sufficient, we still need to decide whether we'll offer a built-in migration mechanism, or just give them a separate withdraw-from-v1-&-deposit-into-v2 proxy function (without any special admin powers).  I think our discussion about this in #25 and on Telegram has sketched some good alternatives on that front.

3. That said - all this raises the question (that we were already discussing, and will guaranteed come up again) of how far we want to go with admin powers.  If we have the ability to switch oracles  then that introduces the question of who will hold the admin keys, will we use multisig, will there be late-night stressful debates between us about whether the recent poor performance of median component X justifies switching to the other oracle, etc...\
\
My point being not that these debates are necessarily bad.  If our oracle breaks I hope we _are_ having late-night discussions about how to fix it!  (For some value of "we", and of "fix".)  But, the ethos of this project from the start (and heck, of cryptocurrencies) was to **try to minimize these debates rather than master them.**  I think we should try to approach all such "admin powers" design questions with this ethos in mind.\
\
Often the answer I'll advocate for will be "If (immutable) v1 is performing poorly, we should try to get out a better (immutable) v2 reasonably quickly, and encourage people to migrate.  This is not as convenient for users as if we had full powers, a support desk, could migrate them ourselves etc: but it protects them by not giving us powers we could abuse, and is the crypto way."\
\
Or, shorter: we're trying to build an excellent vending machine, not an excellent shop.

4. With all that out of the way - I think the misbehaving-oracle scenario you're sketching here is very realistic, and there are governance-minimized ways we can address it!  Thanks for raising it.  What do you think of my `oracle1`/`oracle2` tweak above?

alexroan:
I've altered the description because I think we got some wires crossed here. This is a 0-fee upgrade function that addresses #25 in the event of any failure or even just an upgrade from the team.

1. I think that changes the protocol quite a bit. I dont think we should add more admin powers.
2. This is a PR to resolve #25, and is not specific to oracle failure, but any failure (or even just an upgrade with 0 fees)
3. This only adds admin power to add an optional upgrade address, nothing else.
4. I dont think oracle1/2 is necessary.

jacob-eliosoff:
Ah my apologies for not reading closely enough @alexroan!  Having reread - this was totally my mistake.

Addressing your actual proposal:

There are actually several overlapping risks to be navigated here 😂  So yeah partly that's a reason to just give up on the whole migration feature...  But I'm actually quite sure we could do it safely, just takes some care.

1. The simplest approach is just "50 v1 USM -> 50 v2 USM."  But this is only safe if the v1 ETH price is robust.  And of course v1's price being off is one of the most likely reasons we would launch a v2!  We can't allow a broken v1 USD peg to break v2's peg too.

2. OK, so next simplest is the approach you took here: "Let v2 decide how much ETH a v1 migrator gets out."   This has the vulnerability I described [here](https://github.com/usmfum/USM/issues/25#issuecomment-763402989): anyone who gets access to the keys to set `upgradeAddress` can drain the v1 contract.  We don't want the keys to be that powerful.

3. Arguably next simplest is "migration = a normal v1 `burn`/`defund` at v1's price, *except 0-fee,* combined with a normal v2 `mint`/`fund` at v2's price, also 0-fee."  But 0-fee transactions are fundamentally risky because they mean a temporarily inaccurate price can be exploited by a huge trade.  An attacker could spend money to manipulate the v1 price, do a huge 0-fee migration at the bogus price, and make enough profit to outweigh the manipulation cost.  This is precisely why we added sliding fees (aka larger trades pay more).

4. So here's an example of what's left once we dodge all these risks - apart from the baseline case of "no special migration mechanism":  Do #&#x2060;3 above, but instead of **zero** fee, make the operations **half** fee - so migrators don't get a free ride, but do get a nice discount over crude withdraw-and-redeposit migration.  Halving the fees is not hard but also not totally trivial - one way would be to make `ethFromBurn()` and `ethFromDefund()` accept an additional `feeFraction` argument, which is normally 1, but would be 0.5 when called by the migration fn.  I could flesh this out...  But first let me know if this makes sense.\
\
(Fortunately the migration doesn't need to calculate the half-fee-or-whatever v2 operation: it can leave that for the v2 contract to implement.)

alexroan:
Staying with this proposal,

If the main concern is access to the keys, how much does a multi-sig mitigate that?

There's a Propose -> Confirm flow in here to set the upgrade address, so if we use a multi-sig as well, a majority (or even every) party in the multi-sig has to sign both steps before that address can be used to upgrade.

jacob-eliosoff:
Sure, multisig would help.  To be clear what I'm describing isn't *unsafe* per se - lots of projects are protected by less!  But unless we absolutely need to, I think we should pick:

> "If someone gets the key(s), they can specify a bogus `upgradeAddress`, and try to trick users into migrating to it"

Over:

> "If someone gets the key(s), they can immediately drain the v1 contract"

alexroan:
Good point, the trade-off is that we would need to move quickly if it's an oracle issue. A normal upgrade that should work fine.

I'll adjust the PR.

jacob-eliosoff:
So I was poking at this some more.  There are some really interesting failure modes and on balance I'm guessing we won't include this feature right now...  That said, even if we don't, it is a useful feature and we or someone will sometime!  So I'm curious how to make it work.

1. I see three safety checks we need:
   1. The immediate `msg.sender` must be our specified v2 `upgradeAddress`.  You already check this.
   2. The v1 contract, not v2, must calculate how much ETH is migrated to v2 for a given qty of USM or FUM.  Otherwise we're exposed to the risk I described above: any attacker who gets the keys to set `upgradeAddress` can create a fake v2 contract that uses a "migrate my 1 USM" request to drain the entire v1 ETH pool.  I think implementing this in v1 is pretty easy (basically call `ethFromBurn()`/`ethFromDefund()` with half fees) - sitting down to do so is what spurred me to write this whole comment.
   3. The initiator of the whole migrate/`requestSwap()` operation must be the `holder` whose tokens are being migrated - or at least, we need to check somehow that they've signed the transaction.  Otherwise we're exposed to a similar risk to above: attacker sets `upgradeAddress` to their fake v2 contract and makes it "migrate" *everyone's* ETH into the attacker's pocket.

   **How do we handle #&#x2060;iii?**  I was thinking we could check `holder == tx.origin`, but `tx.origin` seems to be frowned upon or something...  I'm guessing there's another nice way to ensure `holder` has signed off on the operation (or require them to do so)?  (@albertocuestacanada?)  Maybe you're already checking this and I missed it.

   Upon reflection I do think a mechanism where an attacker getting the admin keys lets them drain the v1 contract is a no go.  That level of admin powers goes against the spirit of the project.

2. Because there are (and I think must be) fees/price moves involved, the user may legitimately want to migrate only partially at a time, to save fees.  So we'd probably want separate `requestMigrateUsm(holder, usmQuantity)` and `requestMigrateFum(holder, fumQuantity)` functions.

alexroan:
Good points. Let's reconvene on this after v1!

PR #90, opened 2021-01-23
Upgrade to solc 0.8 and off OpenZeppelin

albertocuestacanada:
Upgraded to solidity 0.8 to get rid of SafeMath.
Changed imports to use my own ERC20, which is simpler than the OpenZeppelin one (and cheaper on gas, marginally).
Cleaned other imports, to reduce the deployment cost.


albertocuestacanada:
Now part of #93

PR #93, opened 2021-01-31
Blacklist feature

jacob-eliosoff:
Have a look, web3/array syntax cost me some tears but it ended up pretty short and clean.  See the two little added tests in test 3 that show USM/FUM sends to the oracle successfully blacklisted.  (We don't actually need to blacklist any addresses in v1, this is planning for the future.)

I think this is ~good to merge


albertocuestacanada (approved):
Simplified significantly the feature.

Sorry for having to merge with `feat/upgrade-08`. I battled for quite a while with a weird eror, and this fixed it.

contracts/USM.sol
```diff
@@ -277,10 +278,10 @@ contract USM is IUSM, Oracle, ERC20Permit, Delegable {
              */
             if (adjustment > 1) {
                 // max(1, old buy price / new mid price):
-                adjustment = WAD.wadMax((uint(storedPrice.value)).mul(adjustment).div(price));
+                adjustment = WAD.wadMax((uint(storedPrice.value)) * (adjustment / price));
             } else if (adjustment < 1) {
                 // min(1, old sell price / new mid price):
-                adjustment = WAD.wadMin((uint(storedPrice.value)).mul(adjustment).div(price));
+                adjustment = WAD.wadMin((uint(storedPrice.value)) * (adjustment / price));
```
jacob-eliosoff, line 284: As above - just take out that last pair of parens.

test/03_USM.test.js
```diff
@@ -651,6 +651,19 @@ contract('USM', (accounts) => {
             ethPool.should.be.bignumber.lt(ethPool0)
           })
 
+          it("users can opt out of receiving FUM", async () => {
```
jacob-eliosoff, line 654: Yeah everything you did here is much better

contracts/USM.sol
```diff
@@ -141,7 +141,7 @@ contract USM is IUSM, Oracle, ERC20Permit, Delegable {
             storedPrice.value = uint224(price);
 
             require(adjustment <= UINT224_MAX, "adjustment overflow");
-            storedBuySellAdjustment.timestamp = uint32(now);
+            storedBuySellAdjustment.timestamp = uint32(block.timestamp);
```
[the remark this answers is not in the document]
jacob-eliosoff: (I see we're not in fact mixing them!  100% `block.timestamp` now.  So fine I guess)

jacob-eliosoff:
Reviewing the way we ended up using `ERC20WithOptOut`, I wonder if maybe it should just be a `WithOptOut` contract that `USM` inherits along with `ERC20Permit`?  That is, our functionality in `ERC20WithOptOut` is actually completely independent from the ERC-20 functionality.  (Only the way we use `noOptOut()` in `USM` combines them.)

This isn't a practical problem, just potentially a slightly cleaner way to organize the code.

albertocuestacanada:
If it's cleaner go for it. I added the functionality in `ERC20Permit` because it made sense and for reusability, but we don't need it to be there.

jacob-eliosoff:
Yeah I think it's one of those things that made sense at the time, but by the time we finished the `ERC20`-`WithOptOut` connection was no longer needed.  I'll merge it

PR #94, opened 2021-01-31
Label versions, eg "Minimal USD v1"

jacob-eliosoff:
Would avoid some user snafus in Etherscan etc...  Also maybe "Minimal**ist** USD v1"?  And should we include it in the symbol too, eg "USMv1"?  Probably not I think but at least in the name.

image


albertocuestacanada:
I'm not sure if Etherscan picks that up from the contract in every instance, we might have to talk to their support team to get it to work.

jacob-eliosoff:
OK but I figure we might as well agree on what we'd *like* to be in there, and pass that to the constructor.  I'm open to suggestions, just don't want a bunch of different versions & testnet deployments showing the exact same symbol + name

PR #97, opened 2021-02-06
Support for a backup oracle.  (Admin-only switching not yet implemented)

jacob-eliosoff:
This is not yet ready to merge (and also I'm not sure we even want this feature), just a draft to work from if we do decide to.

### DIFF
--- a/contracts/USM.sol
+++ b/contracts/USM.sol
@@ -29,6 +29,7 @@ contract USM is IUSM, Oracle, ERC20WithOptOut, Delegable {
     using Address for address payable;
     using WadMath for uint;
 
+    event OracleChanged(address previous, address latest);
     event UnderwaterStatusChanged(bool previous, bool latest);
     event BuySellAdjustmentChanged(uint previous, uint latest);
     event PriceChanged(uint previous, uint latest);
@@ -42,7 +43,9 @@ contract USM is IUSM, Oracle, ERC20WithOptOut, Delegable {
     uint public constant BUY_SELL_ADJUSTMENT_HALF_LIFE = 1 minutes;     // Solidity for 1 * 60
 
     FUM public immutable fum;
-    Oracle public immutable oracle;
+    Oracle public immutable oracle0;        // Primary oracle
+    Oracle public immutable oracle1;        // Backup oracle
+    Oracle public oracle;                   // Currently active oracle (either oracle0 or oracle1)
 
     struct StoredState {
         uint32 timeSystemWentUnderwater;    // Time at which (we noticed) debt ratio went > MAX, or 0 if it's currently < MAX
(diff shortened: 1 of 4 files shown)


contracts/USM.sol
```diff
@@ -67,13 +70,27 @@ contract USM is IUSM, Oracle, ERC20WithOptOut, Delegable {
...
-        oracle = oracle_;
+        oracle0 = oracle0_;
+        oracle1 = oracle1_;
+        oracle = oracle0_;
         fum = new FUM(this, optedOut_);
     }
 
+    // ____________________ Admin-only external transactional functions ____________________
+
+    function setOracle(uint oracleIndex) external override {
```
alexroan, line 84: `onlyOwner()` modifier missing here?

contracts/USM.sol
```diff
@@ -67,13 +70,27 @@ contract USM is IUSM, Oracle, ERC20WithOptOut, Delegable {
...
+        // TODO: need to restrict calls to this fn to admins only!
```
alexroan, line 85: If we switch to the backup, is it likely we'd switch back in future? Unlikely IMO. If we make this function one-way, we could save some gas in a few steps.

If we're keen on keeping this interchangeable, why not go the whole hog and offer a `setOracle(address newOracle)` and have the only admin power be to set a brand new oracle 🤷

contracts/USM.sol
jacob-eliosoff: Yup, I left it out because I wasn't sure how best to implement it, multisig etc...  We'd def need to add `onlyOwner()` (or something) before merging.

contracts/USM.sol
jacob-eliosoff: Because then anyone who gets the admin keys can set a bogus oracle that fixes the ETH price at $0.01, and use a `burn(100 USM)` call to withdraw 10,000 ETH.  There's a huge security difference between "admin can switch between two preset oracles" and "admin can set an arbitrary oracle."

Whether the ability to switch back is useful depends whether you think the median adds any meaningful protection.  That's debatable but clearly I think it does...

I don't think gas cost is a factor for 1-way vs 2-way: the switch is meant to be done only in extreme circumstances anyway, and they both give exactly the same gas cost for `mint()` etc.  It is worth noting though that either a 1-way or 2-way backup does add a cost to `mint()` etc over _no_ backup, because making `oracle` mutable instead of `immutable` adds 736 gas to those fns.  (I guess loading an immutable var is cheaper, kinda makes sense.)

One point in favor of 1-way is, if our primary oracle is really broken (eg 2/3 median components broken), then the ability to switch back to it becomes exploitable as in my $0.01 case above.  Supposing that happens - we switch to backup, and primary seems permanently borked, so that switching back to primary would be exploitable - is there a clean way for us to (provably) burn the admin keys so no one can ever use them again?

jacob-eliosoff:
After some great TG discussion I think we're going to hold off on this feature for now.  If the v1 oracle craps out - time for v2!

PR #98, opened 2021-02-06
Fix emitting of underwater-status-changed events, and also tweak other event emitting

jacob-eliosoff:
1. I was mistakenly emitting a `PriceChanged` event when the underwater status changed - wrong event.  Fixed.
2. In general, do we want to emit events every time the relevant function is called, even if the state var ends up unchanged (previous behavior); or only when state *changes* (behavior I introduce here)?  Fwiw the new code reduces test 3's average `mint()` gas consumption by 1.3k, due to sometimes skipping events (not sure we'd save that much on a typical production call).
3. I'm emitting `timeSystemWentUnderwater`, a timestamp.  I could instead emit a boolean, whether the system is underwater - less information but easier for event-readers to understand.

### DIFF
--- a/contracts/USM.sol
+++ b/contracts/USM.sol
@@ -29,7 +29,7 @@ contract USM is IUSM, Oracle, ERC20WithOptOut, Delegable {
     using Address for address payable;
     using WadMath for uint;
 
-    event UnderwaterStatusChanged(bool previous, bool latest);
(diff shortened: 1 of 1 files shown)


albertocuestacanada:
Events should be emitted to reflect state changes.

Sometimes, events are also sent to enable easier analytics, but that's rare and I don't think we have a need for that.

PR #99, opened 2021-02-07
ERC20WithOptOut -> WithOptOut?

jacob-eliosoff:
See my [comment](https://github.com/usmfum/USM/pull/93#issuecomment-774623207) in #93.  Doesn't seem too important but if people agree this is cleaner let's do it


albertocuestacanada:
Nice and clean :+1:

PR #104, opened 2021-02-09
tweak: defund from fum

albertocuestacanada:


contracts/USM.sol
```diff
@@ -72,11 +72,21 @@ contract USM is IUSM, Oracle, ERC20Permit, WithOptOut, Delegable {
         WithOptOut(optedOut_)
     {
         oracle = oracle_;
-        fum = new FUM(this, optedOut_);
+        FUM fum_ = fum = new FUM(this, optedOut_);
+        addDelegate(address(fum_));
```
jacob-eliosoff, line 76: Do you need this *and* the `msg.sender == address(fum)` condition below?

jacob-eliosoff (changes requested):
Up to you...  If it seems safe I do think this cleans up our interface slightly.  But if we're in any doubt let's leave it as is.  Thanks though sir

contracts/USM.sol
albertocuestacanada: The constructor needs to be reverted, sorry.

contracts/USM.sol
```diff
@@ -121,25 +131,9 @@ contract USM is IUSM, Oracle, ERC20Permit, WithOptOut, Delegable {
...
-     * @notice Defunds the pool by redeeming FUM from an arbitrary address in exchange for equivalent ETH from the pool.
-     * Called only by the FUM contract, when FUM is sent to it.
-     * @param from address to deduct the FUM from.
-     * @param to address to send the ETH to.
-     * @param fumToBurn Amount of FUM to burn.
-     * @param minEthOut Minimum accepted ETH for a successful defund.
-     */
-    function defundFromFUM(address from, address payable to, uint fumToBurn, uint minEthOut)
-        external override
+        onlyHolderOrDelegateOrFUM(from, "Only holder or delegate")
```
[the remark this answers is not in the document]
albertocuestacanada: Yeah, right

albertocuestacanada:
If you like it we can use it, it's safe.

PR #108, opened 2021-02-14
Simplify OurUniswapV2TWAPOracle constructor's args a bit

jacob-eliosoff:
This is not an important change, so we can scrap it if it seems like too much at this stage.  It does make things slightly simpler though while changing basically no code (aside from comments) outside of the `OurUniswapV2TWAPOracle` constructor (and code that uses it).

### DIFF
--- a/contracts/mocks/MockMedianOracle.sol
+++ b/contracts/mocks/MockMedianOracle.sol
@@ -13,10 +13,10 @@ contract MockMedianOracle is MedianOracle, SettableOracle {
     constructor(
         AggregatorV3Interface chainlinkAggregator,
         UniswapAnchoredView compoundView,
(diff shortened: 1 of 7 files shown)


albertocuestacanada (commented):
Looks good to me. I deployed an instance of the MedianOracle with three mock contracts underneath to kovan at `0x5Fcebb9FeB625fE17651dD11F9DC17C8d0eAB63E`. Please kick it a bit to make sure it works right, and if it does, merge.

jacob-eliosoff:
I had a look at your kovan deployment (thanks!) and apart from one little bug in `MockChainlinkAggregatorV3` (so, not mainnet code), which I just fixed in #110, looks good.  Merging.  (I know I'm merging in the opposite order you recommended but I think I can make this work, hang on...)

Issue #107, opened 2021-02-11
Change OurUniswapV2TWAPOracle.latestPrice() to return cached price, not freshly calculated one


jacob-eliosoff:
This will help incentivize anyone using the TWAP to call `refreshPrice()` (rather than `latestPrice()`) in order to get a fresh price, which is beneficial for the other users of the TWAP (like us!).  Even better for us would be to remove `latestPrice()` from the TWAP oracle entirely, but having no `view` function to get a price seems a little harsh.  I think "You can call our `view` function (`latestPrice()`) rather than our transactional fn (`refreshPrice()`), but it won't give you as fresh a price" is a reasonable compromise.

(This is based on a suggestion from @albertocuestacanada last month.  I already incorporated it into `USM`/`USMView`, but I realized tonight I didn't incorporate it into `OurUniswapV2TWAPOracle`.)

Note that this issue has no direct impact on `USM`, which only ever calls `refreshPrice()` anyway.  But it does indirectly affect `USM` in that the more often _other_ users of our oracle call `refreshPrice()`, the more up-to-date the numbers it gives us will be.

jacob-eliosoff:
Resolved by #109.

PR #109, opened 2021-02-14
Change OurUniswapV2TWAPOracle.latestPrice() to *only* update when refreshPrice() is called?

jacob-eliosoff:
This implements #107.

Pros:
1. I think it produces better incentives for other teams that may want to share our `MedianOracle`.  With the old code, they can call `latestPrice()` (`view`, less gas) and get a decently up-to-date price, but their usage doesn't help us at all.  With this change, any user who wants an up-to-date TWAP needs to call `refreshPrice()`, so that the more people/contracts use it, the more up-to-date it is for everyone.

Cons:
1. We don't really _need_ this change.  It might help make our TWAP more up-to-date (a significant vulnerability of the system), but it might not (if no one else reuses our oracle contract), and we can probably live without it.
2. It adds ~230 gas to the cost of each op (`mint` etc).  At least in our tests - in real life it might add more or less, it depends how often `refreshPrice()` gets called vs `latestPrice()`.  (It saves some corresponding gas for `latestPrice()`, but our ops don't actually call `latestPrice()`, only `refreshPrice()`.)

### DIFF
--- a/contracts/mocks/MockChainlinkAggregatorV3.sol
+++ b/contracts/mocks/MockChainlinkAggregatorV3.sol
@@ -2,13 +2,15 @@
 pragma solidity ^0.8.0;
 
 contract MockChainlinkAggregatorV3 {
-    int public _answer;
+    int public price;
+    uint public timestamp;
 
-    function set(int value) external {
-        _answer = value;
+    function set(int price_, uint timestamp_) external {
+        price = price_;
+        timestamp = timestamp_;
     }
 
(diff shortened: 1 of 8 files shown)


albertocuestacanada:
This seems like a very reasonable change to me. 230 gas is 6 cents (!) at current bubble prices so our users are not _that_ inconvenienced.

In exchange there is some incentive for other projects to use our oracle (the more people that use it, the better it is), so it could make a snowball effect to end giving us a really good oracle.

I would do this change, and then shill our oracle to death.

albertocuestacanada (approved):

PR #116, opened 2021-03-12
Final (?) updates to the fee math

jacob-eliosoff:
Three broad changes here, split out into ten commits:

1. Refinements to the fee math in `usmFromMint()`, `ethFromBurn()`, `fumFromFund()` and `ethFromDefund()`, making all four follow the same coherent approach:

   1. Figure out the `adjChangeFactor` by which to scale both a) the system's ETH/USD mid price (which the adjusted buy/sell price will revert back to in a few minutes), and b) the `bidAskAdjustment` temporarily padded on to the mid.  (We always scale both of these by the same factor.)
   2. Using that `adjChangeFactor`, calculate/estimate the buy/sell price (eg, USM buy price for `mint()`, FUM sell price for `defund()`, etc) the operation will _end_ at.  (It's always easy to calculate the buy/sell price the operation _started_ at.)
   3. Using those starting and ending buy/sell prices, calculate/estimate the output - the `usmOut`, `fumOut` or `ethOut` returned to the user.  The choice of formulas here comes down to a mix of correct incentives (higher % fees for larger operations), simplicity and gas-efficiency:
      - For `burn()`, we can cheaply calculate the relevant integral exactly.
      - For the other three, we use approximations that average the starting and ending buy/sell prices to give an avg buy/sell price, and then simply divide/multiply that avg buy/sell price by the input ETH/USM/FUM amount.

2. A related change: now every buy or sell price is calculated from a single ETH/USD price: the `adjustedEthUsdPrice`, ie, the mid ETH/USD price scaled by the `bidAskAdjustment` if appropriate.  Previously, buy/sell prices were calculated using a mix of the adjusted ETH/USD price and the mid ETH/USD price: I think this was unnecessarily complicated.

3. Split `wadExp(x)`/`wadPow(x, y)` into `wadExpDown()`/`wadExpUp()` and `wadPowDown()`/`wadPowUp()`, to ensure rounding is always against the user.  (These new versions aren't yet heavily tested.  Also this finicky rounding may be overkill anyway - not clear whether it's worth the code complexity.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,13 +4,13 @@ Originally proposed by [@jacob-eliosoff](https://github.com/jacob-eliosoff) in [
 
 ## Networks & Addresses
 
-### USM v0.2 (Kovan testnet)
+### USM v0.3 (Kovan testnet)
 
- - Oracle: [0x4B2D04f6d7721B09024E0d61C1617DF04e4C21ae](https://kovan.etherscan.io/address/0x4B2D04f6d7721B09024E0d61C1617DF04e4C21ae)
- - USM: [0x6dcc375aE1f43c8FA4e13D56E6ec584BeE25A88F](https://kovan.etherscan.io/address/0x6dcc375aE1f43c8FA4e13D56E6ec584BeE25A88F)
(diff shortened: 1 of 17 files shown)


albertocuestacanada (approved):
Impressive work, as usual

PR #123, opened 2021-03-18
Fixed ChainlinkOracle to avoid underflow (reject non-positive prices).

jacob-eliosoff:


albertocuestacanada (approved):
We even had the `TODO` there. Can't believe we didn't search for TODO tags before sending it to the auditors. Tsk, tsk.

PR #127, opened 2021-03-18
Tighten up overflow/underflow checking in ethBuffer().

jacob-eliosoff:


albertocuestacanada (approved):
Looks good to me. All that you want is that your positive numbers are below 2e255-1, so that they can be converted to int.

jacob-eliosoff:
Yeah it's a bit subtle.  Basically the reason this code should be safe is:
- `int` vars range from -2<sup>255</sup> to 2<sup>255</sup>-1
- Therefore, `int(ethInPool)` and `int(usmValueInEth)` are both >= 0 and <= 2<sup>255</sup>-1
- Therefore, `buffer = int(ethInPool) - int(usmValueInEth)` is >= -2<sup>255</sup>-1 and <= 2<sup>255</sup>-1
- Therefore, `buffer` lies within the valid `int` range above (<= the maximum 2<sup>255</sup>-1, and > the minimum -2<sup>255</sup>)

PR #131, opened 2021-03-23
Move (non-implementation-specific) comments from USM to IUSM.  (Maybe?)

jacob-eliosoff:
Seems like comments on the purpose of a function, rather than on a specific implementation, belong in the `IUSM` interface rather than the `USM` implementation?  (See ABDK issue #&#x2060;38.)

### DIFF
--- a/contracts/IUSM.sol
+++ b/contracts/IUSM.sol
@@ -77,8 +77,8 @@ abstract contract IUSM is IERC20 {
 
     /**
      * @notice Total amount of ETH in the pool (ie, in the contract).
-     * @return supply the total supply of FUM.  We can't call `usm.fum().totalSupply()` directly because this `IUSM` interface
-     * doesn't (and shouldn't) know about the `FUM` type.
+     * @return supply the total supply of FUM.  Users of this `IUSM` interface, like `USMView`, need to call this rather than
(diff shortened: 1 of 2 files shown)


albertocuestacanada:
Not a chance. Think of the poor guy trying to understand the code and having to flick between the contract and the interface. Like [this](https://github.com/makerdao/dss/blob/master/src/vat.sol).

jacob-eliosoff:
Hmmm.  But what if there were two implementations, should they each duplicate the (non-implementation-specific) comments?  The natural options for non-implementation-specific comments seem to me a) in interface only or b) duplicated in both interface and implementation.  Not c) in implementation only.

I don't actually think someone reading would find the flipping back and forth here so bad.  Most of these are _reasonably_ clear and the interface is very concise.

And in case I deeply resent the comparison with those MakerDAO wankers.

albertocuestacanada:
Ooof, I'll have you know that I've got projects that I can't disclose here, that I'm coding in daiwanese like MakerDAO, _and I'm loving it._

I'm fine duplicating the comments about what the function does in both the contract and the interface. Comments about how something is achieved should be _inside_ the function, not in the natspec, and therefore not duplicated to the interface.

PR #133, opened 2021-03-25
Upgraded Oracle comments to natspec.

jacob-eliosoff:
Just eyeball my natspec/comments for improvements please...


albertocuestacanada:
These are fine, we don't need to tighten the natspec to the level of a specification, methinks.

albertocuestacanada (approved):

PR #134, opened 2021-03-29
enum Round.Up/Down enum -> bool roundUp.  Simpler (and even saves a bit of gas)

jacob-eliosoff:
I don't remember why I had the enum...  Seems like overkill now


albertocuestacanada (approved):
Nice

PR #135, opened 2021-03-29
Debt ratio should really be +inf, not 0, when the ETH pool is worth 0 and the USM isn't.

jacob-eliosoff:


albertocuestacanada:
Is this something we should test?

jacob-eliosoff:
I think test 03 + our launch (which should be the only time the ETH pool's value is 0) are sufficient tests.  This is not a scenario the system should ever hit.  But it is worth double-checking that we're robust to eg a bad oracle price: I've added issue #137.

PR #136, opened 2021-03-30
Replace (eg) wadMulDown(x, y) and wadMulUp(x, y) with wadMul(x, y, roundUp) - maybe?

jacob-eliosoff:
The idea, from ABDK, is to simplify away `wadMulDown()`, `wadMulUp()`, `wadDivDown()`, `wadDivUp()`, and just rely on `wadMul()` and `wadDiv()` delegating to new fn `divRounded()`.  But this seems to add nontrivial gas, eg 1,600 to `defund()`: I think this is either because it results in divisions by WAD being (unnecessarily) checked, or because of overhead from passing the extra `roundUp` arg around.  On balance I'm inclined to scrap this change and leave `wadMulDown()` etc as they are.

### DIFF
--- a/contracts/IUSM.sol
+++ b/contracts/IUSM.sol
@@ -2,7 +2,6 @@
 pragma solidity ^0.8.0;
 
 import "acc-erc20/contracts/IERC20.sol";
-import "./WadMath.sol";
 
 abstract contract IUSM is IERC20 {
     event UnderwaterStatusChanged(bool underwater);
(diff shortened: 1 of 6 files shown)


albertocuestacanada:
For 1600 gas, I wouldn't do this refactor. It doesn't make the code that much clearer, and it is invisible to users.

jacob-eliosoff:
Agreed.  There may be a way to do it without that gas cost, or the gas hit may be something dumb I did in the change...  But not gonna spend a lot of time on it.

PR #138, opened 2021-03-31
Simplified wadPowDown() to one case by using ints rather than uints.

jacob-eliosoff:
Motivated by ABDK issue #&#x2060;49.  The `wadPowDown()`/`wadPowUp()` tests in test 03 passed on the first try, that's a good sign.  This uses a tiny bit more gas (30-80 per mint/burn op) but the simplicity and avoidance of if-else cases definitely seems worth it.

### DIFF
--- a/contracts/WadMath.sol
+++ b/contracts/WadMath.sol
@@ -195,7 +195,7 @@ library WadMath {
      *
      * @dev We're given X = x * 1e18, and Y = y * 1e18 (both WAD-formatted); we want Z = z * 1e18, where z =~ x**y; and
      * we have `log_2(x)`, which returns log2(x) * 2**121, and `pow_2(X = x * 2**121)`, which returns 2**x = 2**(X / 2**121).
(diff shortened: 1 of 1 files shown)


albertocuestacanada:
This goes over my head, unfortunately.

In principle, I like the more compact code, despite the slightly higher cost. Simpler code leads to less bugs in my opinion.

I'm concerned about overflows, although you seem to have thought about them. I think that Mikhail would catch if there is anything wrong with the code.

As a side note, recent solidity compilers allow you to cast between int and uint, but keeping the byte length, or between byte lengths, but keeping the uint/int. That means that when you want to change both between uint/int and byte lengths, you need two casts.

jacob-eliosoff:
Yeah, just looking at the diffs here, I can't really imagine choosing the old code over the new.  Merging.

The "can't change both bit length and uint-vs-int at once" behavior seems reasonable to me...  The resulting cast just looked weird!

PR #141, opened 2021-03-31
WithOptOut -> OptOutable (maybe?), and add explanatory comment

jacob-eliosoff:
I dunno, I could go either way on the renaming (see ABDK issue #&#x2060;83).  The comment is probably worth adding though


albertocuestacanada:
`OptOutable` is quite an ugly word, but there aren't many decent options. `UserCanOptOut` is another ugly option. I can go any way. The comment is very good, it took me a second to remember what this was about.

albertocuestacanada (approved):

jacob-eliosoff:
Yeah all the name options I've thought of are pretty bad.  But if I was describing this behavior to someone in English, I think "opt-outable" is how I'd describe it, so let's go with that for now - easy to change if we think of something more natural.  Fwiw some poor schmo on the Internet went with the same bad name: https://tighten.co/blog/opting-out-a-simple-solution-for-conditionally-cancelling-laravel-notifications/.

PR #143, opened 2021-04-01
medianIndex() -> medianPriceAndUpdateTime().

jacob-eliosoff:
This is less general median code and a little less elegant, but arguably clearer (more direct about what it's doing) and less redundant (and saves like 50 gas, whatever).

But this is still a bit of an "only maybe" change...


albertocuestacanada (approved):
It's a bit weird that you pass the times to `medianPriceAndUpdateTime` just to return them with the price that you associated them to, but if it saves gas I'm cool with it.

jacob-eliosoff:
It is a bit weird but the most natural alternative I see is returning the median index as in the old code, and as ABDK said, leaving aside gas I think that code is a little more messily redundant than this is.

PR #144, opened 2021-04-01
Scrap Ownable

jacob-eliosoff:


jacob-eliosoff:
This also somehow saves 800+ gas in `fund()` and `defund()` 🙄

albertocuestacanada (approved):
800 gas is an SLOAD, I wonder where did it come from...

PR #145, opened 2021-04-01
Make WithOptOut emit OptOutStatusChanged events

jacob-eliosoff:
Really nbd but hey (ABDK issue #&#x2060;86)


albertocuestacanada (changes requested):
I don't mind following this pattern if Mikhail likes it, but then we should do the same in `Delegable.sol` (which reverts when trying to delegate someone that was a delegate already).

Would you mind changing Delegable.sol to follow this pattern?

jacob-eliosoff:
Yup makes sense.  Updated with the `Delegable` change.

albertocuestacanada (approved):
Thanks!

PR #146, opened 2021-04-02
More clever (and more gas-efficient!) approach to decaying the bid-ask adjustment and minFumBuyPrice

jacob-eliosoff:
Courtesy of ABDK.  Clearer, gets rid of my `wadHalfExpApprox()` hack, higher-precision, _and_ saves ~500 gas per mint/burn operation!  Good stuff.


albertocuestacanada (approved):
We are not worthy of his genius.

Issue #113, opened 2021-03-01
Maybe let user burn() USM even when the system is underwater, just redeeming at < 100%?


jacob-eliosoff:
There are arguments for and against this, but it's a reasonable option to consider.  (Suggested by ABDK audit)

jacob-eliosoff:
Implemented by #147.

PR #147, opened 2021-04-05
Allow USM burns when debt ratio > 100%: just haircut them proportionally

jacob-eliosoff:
I haven't made up my mind whether I think this change is desirable - should we "protect burners from themselves" by disallowing (USM) `burns` until the system is back above water (status quo behavior), or let them burn at a discount if they choose (ABDK's suggestion, #113)?  But anyway, it seems like a simple self-contained change, so we can decide based on first principles rather than implementation difficulty.

### DIFF
--- a/contracts/USM.sol
+++ b/contracts/USM.sol
@@ -211,11 +211,7 @@ contract USM is IUSM, Oracle, ERC20Permit, WithOptOut, Delegable {
         ls.bidAskAdjustment = ls.bidAskAdjustment.wadMulUp(adjGrowthFactor);
         ls.ethUsdPrice = ls.ethUsdPrice.wadMulUp(adjGrowthFactor);
 
-        // 5. Check that the burn didn't leave debt ratio > 100%:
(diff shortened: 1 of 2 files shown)


albertocuestacanada:
Sorry, I need a bit more background, I don't get the issue.

Based on first principles, I would let the users choose to take a loss if they think it's best.

jacob-eliosoff:
This is based on a quick back-&-forth with one of the ABDK guys on Telegram.  The idea is, suppose there's 100,000 USM outstanding but the ETH in the pool is currently only worth $90,000: the system is underwater.
* The old code says, no `burns` (USM redemptions) allowed until the system gets back above water: either because FUM buyers step up to recapitalize the system, or just because the ETH/USD price climbs back up.
* The new code says, sure you can `burn` while underwater, but each USM will only get you back $0.90 worth of ETH, because that's all we have for you.

Both behaviors seem reasonable.  The main reason I originally chose to disallow underwater redemptions is to avoid users accidentally burning at an unfortunate time and getting back less than they could (this would be a point for a nice web UI to ask "Are you sure you want to do this?").  But there is a case for letting the user decide, I dunno.

albertocuestacanada:
Yeah, my vote would be to let users take a loss of they want, instead of locking their assets.

PR #160, opened 2021-10-05
tweak: removed the use of Delegable

alcueca:
Removed the use of Delegable, since it's untested and by now proven cumbersome.

### DIFF
--- a/contracts/USM.sol
+++ b/contracts/USM.sol
@@ -6,7 +6,6 @@ import "./IUSM.sol";
 import "./OptOutable.sol";
 import "./oracles/Oracle.sol";
 import "./Address.sol";
-import "./Delegable.sol";
 import "./WadMath.sol";
 import "./FUM.sol";
 import "./MinOut.sol";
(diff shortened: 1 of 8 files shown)


contracts/USM.sol
```diff
@@ -64,22 +63,11 @@ contract USM is IUSM, ERC20Permit, OptOutable, Delegable {
         OptOutable(optedOut_)
     {
         oracle = oracle_;
-        fum = new FUM(optedOut_);
+        fum = IFUM(address(new FUM(optedOut_)));
```
jacob-eliosoff, line 66: This change doesn't seem necessary?

jacob-eliosoff (changes requested):
Plus remove `Delegable.sol` itself, right?

I can't find anything wrong here but I'm a little nervous about just this many line changes at this stage if the old code was basically working...  I lean in favor of merging but please give this an extra dose of scrutiny given the cost/benefit right now...

